PR #224, opened 2014-07-29
Use --data-dev or --data-dir for ceph-disk parameters to avoid making di...

osynge:
ceph-disk parameters "--data-dev" or "--data-dir" are useful to avoid making directories such as /dev/sdc by mistake when you want to use a device that does not exist.


alfredodeza:
ceph-disk is no longer a part of ceph

PR #254, opened 2014-12-04
Implement the ceph osd destroy feature in ceph-deploy.

Vicente-Cheng:
Feature:
        1. handle osd already out
        2. handle osd did not blong any hostname
        3. handle osd is not in acting set

    Usage: ceph-deploy osd destroy <hostname> --osd-id <osd_id>

    I have tested in CentOS6/7, Ubuntu12.04/14.04 and it work normally.
    This feature can help to remove osd quickly. It maybe is helpful for
    newbie who uses ceph.

    If you have any questions, feel free to let me know.
    Thanks!

    Signed-off-by: Vicente Cheng <freeze.bilsted@gmail.com>

### DIFF
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -10,7 +10,7 @@ from textwrap import dedent
 from cStringIO import StringIO
 
 from ceph_deploy import conf, exc, hosts
-from ceph_deploy.util import constants, system
+from ceph_deploy.util import constants, system, paths
 from ceph_deploy.cliutil import priority
 from ceph_deploy.lib import remoto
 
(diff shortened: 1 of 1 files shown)


ceph_deploy/osd.py
```diff
@@ -388,6 +388,228 @@ def activate(args, cfg):
...
+        LOG.info("(NOT IMPLEMENT) remove all osd")
+        raise NotImplementedError("It need to assign the specific osd id.")
+
+    command = [
+        'ceph',
+        '--cluster={cluster}'.format(cluster=cluster),
+        'osd',
+        'tree',
+        '--format=json',
+    ]
```
trhoden: After looking at this for a bit, I have a few suggestions/requests.  Starting here, I don't think we need to call ceph osd tree explicity here.  Contained in this osd.py source file there is already an osd_tree() function that can be re-used.

I also think it would be worthwhile to look at the code in osd_list().  Specifically the part about gathering the osd tree from a monitor node.  This code, as written, goes to each server hosting the OSD and calls "ceph osd tree".  however, it is possible that OSD nodes do not have the cephx admin key that allows that command to work.  It would be better to gather the osd tree once from a monitor node up in destroy(), before going into the loop.

trhoden:
@Vicente-Cheng I wanted to point you to http://tracker.ceph.com/issues/7454, which has some details about how we ultimately want to roll this capability into Ceph.

A lot of the logic contained in this PR can be in ceph-disk instead.  ceph-deploy uses ceph-disk where it can, so it doesn't have to concern itself with too many low-level disk details.  We would like to remove OSDs via ceph-disk as well, then ultimately have ceph-deploy use the relevant ceph-disk command instead.

This also makes a lot of the logic required to remove an OSD re-usable by other tools besides ceph-deploy.

ceph_deploy/osd.py
Vicente-Cheng: I completely agree your opinion.
It sounds more sensible abut we use the osd_tree() function.

OK, I think I will check osd_list() and osd_tree() first.
Then the better way is that we can gather the osd tree once from **monitor** node when we call destroy() function by osd_list()/osd_tree() ?

Vicente-Cheng:
@trhoden 
OK, I will study this issue and its discussion (w/ related issues) and think how we implement this feature via ceph-disk.

Vicente-Cheng:
@trhoden 
I remove the manual "ceph osd tree" called component.
To complete my goal (get the osd tree) by osd_tree() function.

I plan to give the "ceph-deploy osd" two subcommand like our discussion "deactivate" and "destroy".
The "destroy" will have dependency on "deactivate". how did you think?

As you work on "deactivate", I am going to work on ceph-disk "**destroy**".
When these two subcommand complete, I will modify the ceph-deploy to complete these two features.

dachary:
@Vicente-Cheng closing because I'm assuming your latest contribution to ceph-disk address the same problem. Feel free to re-open if I'm mistaken.

Vicente-Cheng:
@dachary Sure, I will complete the ceph-disk feature first.
Then ceph-deploy could use the ceph-disk related feature to complete it.

PR #256, opened 2015-01-02
rpmlint fixes

ktdreyer:
This pull request fixes two types of rpmlint warnings in the ceph-deploy RPM packaging.

The first commit changes the tabs to spaces in the `ceph-deploy.spec` file.

The second commit quotes the RPM macros in the comments of `ceph-deploy.spec`.


ktdreyer:
Thanks for reviewing/merging!

PR #257, opened 2015-01-06
Add abstracted function about get monitor members.

Vicente-Cheng:
To get monitor members feature will be reuse serval times in mon.py and osd.py.
    Due to this reason, I rewrite the helper function for this feature.

    If you have any questions, feel free to let me know.
    Thanks!

    Signed-off-by: Vicente Cheng <freeze.bilsted@gmail.com>

### DIFF
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -232,8 +232,9 @@ def mon_add(args):
 
 
 def mon_create(args):
-
-    args.mon = get_mon_member(args)
+    cfg = conf.ceph.load(args)
+    if not args.mon:
+        args.mon = get_mon_initial_members(args)
 
     if not args.mon:
         raise exc.NeedHostError()
@@ -395,7 +396,7 @@ def mon_destroy(args):
 
 
 def mon_create_initial(args):
-    mon_initial_members = get_mon_member(args, True)
+    mon_initial_members = get_mon_initial_members(args, error_on_empty=True)
 
     # create them normally through mon_create
     mon_create(args)
@@ -528,25 +529,21 @@ def make(parser):
 # Helpers
 #
 
-def get_mon_member(args, init_tags=False):
+
+def get_mon_initial_members(args, error_on_empty=False):
     """
-    Abstracte the process about getting mon_member.
+    Read the ceph config file and return the value of mon_initial_members
 
-    @init_tags for choose normal/init stage
-
-    return value are monitor members
+    Optionally, a NeedHostError can be raised if the value is None.
     """
     cfg = conf.ceph.load(args)
-    if init_tags is not True:
-        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        monitors = re.split(r'[,\s]+', mon_initial_members)
-        return monitors
-    else:
-        cfg_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        if cfg_initial_members is None:
-            raise RuntimeError('No `mon initial members` defined in config')
-        mon_initial_members = re.split(r'[,\s]+', cfg_initial_members)
-        return mon_initial_members
+    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
+    monitors = re.split(r'[,\s]+', mon_initial_members)
+    if not monitors and error_on_empty:
+        raise exc.NeedHostError(
+            'could not find `mon initial members` defined in ceph.conf'
+        )
+    return mon_initial_members
 
 
 def is_running(conn, args):
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -58,7 +58,7 @@ def catch_mon_errors(conn, logger, hostname, cfg, args):
     and warn apropriately about it.
     """
     monmap = mon_status_check(conn, logger, hostname, args).get('monmap', {})
-    mon_initial_members = get_mon_initial_members(args, error_on_empty=True, _cfg=cfg)
+    mon_initial_members = get_mon_initial_members(args, _cfg=cfg)
     public_addr = cfg.safe_get('global', 'public_addr')
     public_network = cfg.safe_get('global', 'public_network')
     mon_in_monmap = [
@@ -538,12 +538,14 @@ def get_mon_initial_members(args, error_on_empty=False, _cfg=None):
     else:
         cfg = conf.ceph.load(args)
     mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-    if not mon_initial_members and error_on_empty:
-        raise exc.NeedHostError(
-            'could not find `mon initial members` defined in ceph.conf'
-        )
-    monitors = re.split(r'[,\s]+', mon_initial_members)
-    return monitors
+    if not mon_initial_members:
+        if error_on_empty:
+            raise exc.NeedHostError(
+                'could not find `mon initial members` defined in ceph.conf'
+            )
+    else:
+        mon_initial_members = re.split(r'[,\s]+', mon_initial_members)
+    return mon_initial_members
 
 
 def is_running(conn, args):
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -58,7 +58,7 @@ def catch_mon_errors(conn, logger, hostname, cfg, args):
     and warn apropriately about it.
     """
     monmap = mon_status_check(conn, logger, hostname, args).get('monmap', {})
-    mon_initial_members = get_mon_initial_members(args, _cfg=cfg)
+    mon_initial_members = get_mon_initial_members(args, error_on_empty=True, _cfg=cfg)
     public_addr = cfg.safe_get('global', 'public_addr')
     public_network = cfg.safe_get('global', 'public_network')
     mon_in_monmap = [
@@ -538,11 +538,11 @@ def get_mon_initial_members(args, error_on_empty=False, _cfg=None):
     else:
         cfg = conf.ceph.load(args)
     mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-    monitors = re.split(r'[,\s]+', mon_initial_members)
-    if not monitors and error_on_empty:
+    if not mon_initial_members and error_on_empty:
         raise exc.NeedHostError(
             'could not find `mon initial members` defined in ceph.conf'
         )
+    monitors = re.split(r'[,\s]+', mon_initial_members)
     return monitors
 
 
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -58,7 +58,7 @@ def catch_mon_errors(conn, logger, hostname, cfg, args):
     and warn apropriately about it.
     """
     monmap = mon_status_check(conn, logger, hostname, args).get('monmap', {})
-    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
+    mon_initial_members = get_mon_initial_members(args, _cfg=cfg)
     public_addr = cfg.safe_get('global', 'public_addr')
     public_network = cfg.safe_get('global', 'public_network')
     mon_in_monmap = [
@@ -234,7 +234,7 @@ def mon_add(args):
 def mon_create(args):
     cfg = conf.ceph.load(args)
     if not args.mon:
-        args.mon = get_mon_initial_members(args)
+        args.mon = get_mon_initial_members(args, _cfg=cfg)
 
     if not args.mon:
         raise exc.NeedHostError()
@@ -530,20 +530,23 @@ def make(parser):
 #
 
 
-def get_mon_initial_members(args, error_on_empty=False):
+def get_mon_initial_members(args, error_on_empty=False, _cfg=None):
     """
     Read the ceph config file and return the value of mon_initial_members
 
     Optionally, a NeedHostError can be raised if the value is None.
     """
-    cfg = conf.ceph.load(args)
+    if _cfg:
+        cfg = _cfg
+    else:
+        cfg = conf.ceph.load(args)
     mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
     monitors = re.split(r'[,\s]+', mon_initial_members)
     if not monitors and error_on_empty:
         raise exc.NeedHostError(
             'could not find `mon initial members` defined in ceph.conf'
         )
-    return mon_initial_members
+    return monitors
 
 
 def is_running(conn, args):
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -233,10 +233,7 @@ def mon_add(args):
 
 def mon_create(args):
 
-    cfg = conf.ceph.load(args)
-    if not args.mon:
-        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        args.mon = re.split(r'[,\s]+', mon_initial_members)
+    args.mon = get_mon_member(args)
 
     if not args.mon:
         raise exc.NeedHostError()
@@ -398,11 +395,7 @@ def mon_destroy(args):
 
 
 def mon_create_initial(args):
-    cfg = conf.ceph.load(args)
-    cfg_initial_members = cfg.safe_get('global', 'mon_initial_members')
-    if cfg_initial_members is None:
-        raise RuntimeError('No `mon initial members` defined in config')
-    mon_initial_members = re.split(r'[,\s]+', cfg_initial_members)
+    mon_initial_members = get_mon_member(args, True)
 
     # create them normally through mon_create
     mon_create(args)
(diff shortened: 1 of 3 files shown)


ceph_deploy/mon.py
```diff
@@ -535,6 +528,26 @@ def make(parser):
 # Helpers
 #
 
+def get_mon_member(args, init_tags=False):
```
trhoden: I'd prefer this to be named get_mon_members(), since there can me many monitors.

ceph_deploy/mon.py
```diff
@@ -233,10 +233,7 @@ def mon_add(args):
 
 def mon_create(args):
 
-    cfg = conf.ceph.load(args)
-    if not args.mon:
-        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        args.mon = re.split(r'[,\s]+', mon_initial_members)
+    args.mon = get_mon_member(args)
```
trhoden: I think this line still needs to be in a "if not args.mon" block.  It should only be reading the mon_initial_members from the config file if args.mon is not already filled out.

ceph_deploy/mon.py
```diff
@@ -535,6 +528,26 @@ def make(parser):
...
+        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
+        monitors = re.split(r'[,\s]+', mon_initial_members)
+        return monitors
+    else:
+        cfg_initial_members = cfg.safe_get('global', 'mon_initial_members')
+        if cfg_initial_members is None:
+            raise RuntimeError('No `mon initial members` defined in config')
+        mon_initial_members = re.split(r'[,\s]+', cfg_initial_members)
+        return mon_initial_members
+
```
trhoden: I'm a bit confused by the structure of this.  It seems the only difference between the two code paths (whether or not init_tags is True) is whether or not it raises an exception if cfg.safe_get() returns None.  This seems much easier written as:

```python
def get_mon_initial_members(args, error_on_empty=False):
    cfg = conf.ceph.load(args)
    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
    monitors = re.split(r'[,\s]+', mon_initial_members)
    if error_on_empty:
        raise exc.NeedHostError(
            'could not find `mon initial members` defined in ceph.conf'
         )
    return mon_initial_members
```

After looking closer at this, since this function is only to get mon_initial_members from the config file, let's rename it to get_mon_initial_members()

It would be better to raise exc.NeedHostError() rather than the RuntimeError, since that's what was raised in the other places.

ceph_deploy/osd.py
```diff
@@ -9,7 +9,7 @@
 
 from cStringIO import StringIO
 
-from ceph_deploy import conf, exc, hosts
+from ceph_deploy import conf, exc, mon, hosts,
```
trhoden: Can you remove the trailing comma here?

ceph_deploy/mon.py
```diff
@@ -398,11 +395,7 @@ def mon_destroy(args):
 
 
 def mon_create_initial(args):
-    cfg = conf.ceph.load(args)
-    cfg_initial_members = cfg.safe_get('global', 'mon_initial_members')
-    if cfg_initial_members is None:
-        raise RuntimeError('No `mon initial members` defined in config')
-    mon_initial_members = re.split(r'[,\s]+', cfg_initial_members)
+    mon_initial_members = get_mon_member(args, True)
```
trhoden: Since the second argument to get_mon_member is a kwarg, I'd rather see this as get_mon_member(args, init_tags=True).

ceph_deploy/mon.py
```diff
@@ -535,6 +528,26 @@ def make(parser):
...
 #
 
+def get_mon_member(args, init_tags=False):
+    """
+    Abstracte the process about getting mon_member.
+
+    @init_tags for choose normal/init stage
+
+    return value are monitor members
+    """
```
trhoden: I really appreciate the documentation of input params and return values.  However, it's a bit inconsistent with the rest of the functions in ceph-deploy.  There is no where else that follows this style, especially with the @ symbol.

I'd rather just see a simple comment about what the function does, like:
```
Read and return the value of mon_initial_members from the Ceph config file

Optionally, a NeedHostError exception can be raised if the value is None.
```

ceph_deploy/osd.py
```diff
@@ -466,11 +466,8 @@ def disk_list(args, cfg):
...
-    # FIXME: this portion should probably be abstracted. We do the same in
-    # mon.py
-    cfg = conf.ceph.load(args)
-    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-    monitors = re.split(r'[,\s]+', mon_initial_members)
+
+    monitors = mon.get_mon_member(args)
 
     if not monitors:
         raise exc.NeedHostError(
```
trhoden: If you pass the flag to raise an exception on error to get_mon_members, the additional check and raise can be removed.

ceph_deploy/mon.py
```diff
@@ -232,11 +232,9 @@ def mon_add(args):
 
 
 def mon_create(args):
-
     cfg = conf.ceph.load(args)
     if not args.mon:
-        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        args.mon = re.split(r'[,\s]+', mon_initial_members)
+        args.mon = get_mon_initial_members(args)
```
Vicente-Cheng: @trhoden 
I save the cfg variable because the catch_mon_errors() would use it in later line _**276**_ .
This will make us to call conf load twice in the mon_create().
one in the first, another in the get_mon_initial_members().

If we pass the _args_ to catch_mon_errors(), can we do the config load in catch_mon_error()?

ceph_deploy/osd.py
```diff
@@ -466,16 +466,8 @@ def disk_list(args, cfg):
...
-    # mon.py
-    cfg = conf.ceph.load(args)
-    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-    monitors = re.split(r'[,\s]+', mon_initial_members)
 
-    if not monitors:
-        raise exc.NeedHostError(
-            'could not find `mon initial members` defined in ceph.conf'
-        )
+    monitors = mon.get_mon_initial_members(args)
```
trhoden: This call needs error_on_empty=True added to it.

ceph_deploy/mon.py
trhoden: I see what you are talking about.

Here's what I would do...

Add _cfg=None to the method signature get_mon_initial_members().  That way you can pass in a load cfg if you already have one.  if not, make it load one.  I'll populate the rest of the comments to that effect.

ceph_deploy/mon.py
```diff
@@ -536,6 +530,22 @@ def make(parser):
 #
 
 
+def get_mon_initial_members(args, error_on_empty=False):
```
trhoden: Make this get_mon_initial_members(args, error_on_empty=False, _cfg=None):

ceph_deploy/mon.py
```diff
@@ -536,6 +530,22 @@ def make(parser):
...
+    """
+    Read the ceph config file and return the value of mon_initial_members
+
+    Optionally, a NeedHostError can be raised if the value is None.
+    """
+    cfg = conf.ceph.load(args)
```
trhoden: cfg = _cfg or conf.ceph.load(args)

ceph_deploy/mon.py
```diff
@@ -232,11 +232,9 @@ def mon_add(args):
 
 
 def mon_create(args):
-
     cfg = conf.ceph.load(args)
     if not args.mon:
-        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        args.mon = re.split(r'[,\s]+', mon_initial_members)
+        args.mon = get_mon_initial_members(args)
```
trhoden: args.mon = get_mon_initial_members(args, _cfg=cfg)

ceph_deploy/mon.py
trhoden: Please also use the new abstracted function in catch_mon_errors(), line *61*

Vicente-Cheng:
Hi @trhoden 
I already add the new parameter _cfg and use the new abstracted function in catch_mon_errors().

Test OK in CentOS7 and ubuntu12.04

Thanks!!
vicente

ceph_deploy/mon.py
```diff
@@ -232,11 +232,9 @@ def mon_add(args):
...
 def mon_create(args):
-
     cfg = conf.ceph.load(args)
     if not args.mon:
-        mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
-        args.mon = re.split(r'[,\s]+', mon_initial_members)
+        args.mon = get_mon_initial_members(args, _cfg=cfg)
 
     if not args.mon:
         raise exc.NeedHostError()
```
trhoden: If we pass error_on_empty=True to get_mon_intial_members() just above this, this raise can go, correct?  Looks that way to me.

ceph_deploy/mon.py
```diff
@@ -536,6 +530,25 @@ def make(parser):
...
+def get_mon_initial_members(args, error_on_empty=False, _cfg=None):
+    """
+    Read the ceph config file and return the value of mon_initial_members
+
+    Optionally, a NeedHostError can be raised if the value is None.
+    """
+    if _cfg:
+        cfg = _cfg
+    else:
+        cfg = conf.ceph.load(args)
```
trhoden: Just a nit - this can be done with
`cfg = _cfg or conf.ceph.load(args)`

You don't need to change it thought.

trhoden:
@Vicente-Cheng Other than the one last change (a `raise` I don't think is needed any more), this is looking good!

Vicente-Cheng:
Hi @trhoden 
Your opinion is right.
We can pass the *error_on_empty=True* and **raise** is not needed any more.
I improve it.

also TEST OK in CentOS7 and ubuntu12.04

Thanks!!
vicente

ceph_deploy/mon.py
Vicente-Cheng: That let me know a new method to handle it!

Thanks for your reminding :+1:

trhoden:
OK to test

trhoden:
test this please

Vicente-Cheng:
Hi @trhoden 

I change the if condition before we get the monitors.
That can handle None String input.

Could you make jenkins to test it again?

Thanks!!
vicente

trhoden:
test this please

Vicente-Cheng:
Hi @trhoden 

Sorry about that I do not understand this test failed.
(Only see the abort message and see older result in the next test #1444)
makes me confused :P 

thanks for your answering!!

vicente

alfredodeza:
test this please

alfredodeza:
retest this please

Vicente-Cheng:
Hi @trhoden @alfredodeza :

I check the unit test function.
The previous failure due to the test catch_mon_errors()

TEST will pass None into catch_mon_errors() but re.split can not handle this.
I try to add error_on_empty=True in catch_mon_errors() then will get the raise error.
(It is right but the unit test function failure due to this raise)

Now I hold the error_on_empty=False when catch_mon_errors() call.
Then, return None if mon_initial_members is None.
(It means give the handling back to catch_mon_errors() if mon_initial_members is None.)

I try to run unit test. It looks fine.

Would you have any suggestions about this change?
(I try to change the unit test but I could not find the suitable method.)
Or we can make jenkins test again?

thanks!!
vicente

Vicente-Cheng:
HI @trhoden 

For previous failure.
I modified the test_mon.py function for catch the raise and did not modify the abstract function.

I have tested the unittest. It looks fine.
Would you take some time to this commit?

If you have any opinions, feel free to let me know!

very thanks!!
vicente

ceph_deploy/mon.py
```diff
@@ -58,7 +58,7 @@ def catch_mon_errors(conn, logger, hostname, cfg, args):
     and warn apropriately about it.
     """
     monmap = mon_status_check(conn, logger, hostname, args).get('monmap', {})
-    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
+    mon_initial_members = get_mon_initial_members(args, error_on_empty=True, _cfg=cfg)
```
trhoden: I don't think this one should have error_on_empty=True.  The follow-on logic is counting on mon_initial_members being `None` if there are no members.

trhoden:
Hi Vicente,

There are a few things that should happen here.

New unit tests should be added in tests/test_mon.py to test the new get_mon_initial_members() function. This should include tests to make sure that the function correctly returns a list with one item when there is only one in the config, or a list of multiple items when you add more than one -- this will test the splitting of a comma separated list in the config.

It should also test that when there is no value in the config, that an exception is raised when error_on_empty is True, and that None is returned when False.  You can use 'with pytest.raises' to ensure that exceptions are raised when needed.  Checkout the Py.test docs.

As for the existing tests for mon.catch_mon_errors(), I actually don't think these will need to be modified.  Instead, I think the problem is that error_on_empty is to to `True` in catch_mon_errors().  I think that is a mistake, and it may be something I suggested a while ago.  My memory is failing.  :)

I would make sure that is not set, and I would try re-running with the tests back they way they were.  I would like to see unit tests for the new function, though.

ceph_deploy/mon.py
```diff
@@ -536,6 +527,25 @@ def make(parser):
...
+    if _cfg:
+        cfg = _cfg
+    else:
+        cfg = conf.ceph.load(args)
+    mon_initial_members = cfg.safe_get('global', 'mon_initial_members')
+    if not mon_initial_members and error_on_empty:
+        raise exc.NeedHostError(
+            'could not find `mon initial members` defined in ceph.conf'
+        )
+    monitors = re.split(r'[,\s]+', mon_initial_members)
```
trhoden: I think this is going to throw an exception if error_on_empty is False.  The pre-ceding code will throw an expection if error_on_empty is True, but if error_on_empty is False, `mon_initial_members` could be None.  If it is, this is going to call re.split(...., None), and that will lead to a TypeError.  You can't give None to split()

Instead, this needs to be:

```
if not mon_initial_members:
    if error_on_empty:
         raise...
else:
    mon_initial_members = re.split(r'[,\s]+', mon_initial_members)
return mon_initial_members
```
There is no need for the `monitors` intermediate variable. I feel like it was like that at one point, but got changed.  I could be wrong.

trhoden:
Once all that is working, I think this can all be rebased/squashed into one or two commits at most.  1 commit for the changes to the functional code, and a second commit for the test additions.

Vicente-Cheng:
hi @trhoden 

I add the new unit test function for get_mon_initial_members().
Test function include single, multiple, none item input and raise assertion.

Change the catch_mon_errors() to previous flow.

Also tested. It looks fine.

I will rebase into one or two commit when we reach a consensus.

thanks!!
vicente

trhoden:
Hey @Vicente-Cheng, this is looking a lot better.  Thanks for sticking with it!  I've got a few line comments for you, but that's about it...

ceph_deploy/tests/test_mon.py
```diff
@@ -20,6 +21,45 @@ def make_fake_conn(receive_returns=None):
...
+        cfg = make_fake_conf()
+        with pytest.raises(exc.NeedHostError):
+            mon.get_mon_initial_members(Mock(), True, cfg)
+
+    def test_return_if_mon_none_and_empty_false(self):
+        cfg = make_fake_conf()
+        mon_initial_members = mon.get_mon_initial_members(Mock(), False, cfg)
+        if mon_initial_members is None:
+            return True
+        raise AssertionError("Return value is invalid! (Should be None)")
```
trhoden: Normally with unit tests you would just assert that the return is what you know it should be.

So I think this can just be:

assert mon_initial_members == None

It's actually unusual to raise your own AssertionError in tests.  You instead assert what you want the test to result it.

http://pytest.org/latest/assert.html may help

ceph_deploy/tests/test_mon.py
```diff
@@ -20,6 +21,45 @@ def make_fake_conn(receive_returns=None):
...
+
+    def test_single_item_if_mon_not_none(self):
+        cfg = make_fake_conf()
+        cfg.add_section('global')
+        cfg.set('global', 'mon initial members', 'AAAA')
+        mon_initial_members = mon.get_mon_initial_members(Mock(), False, cfg)
+        self.list_and_size_checker(mon_initial_members, 'single')
```
trhoden: I think this can be "assert mon_initial_members == ['AAAA']" ?

ceph_deploy/tests/test_mon.py
```diff
@@ -20,6 +21,45 @@ def make_fake_conn(receive_returns=None):
...
+
+    def test_multiple_item_if_mon_not_none(self):
+        cfg = make_fake_conf()
+        cfg.add_section('global')
+        cfg.set('global', 'mon initial members', 'AAAA, BBBB')
+        mon_initial_members = mon.get_mon_initial_members(Mock(), False, cfg)
+        self.list_and_size_checker(mon_initial_members, 'multiple')
```
trhoden: assert set(mon_initial_members) == set(['AAAA', 'BBBB']). 

In fact, you could use the set conversion for both comparisons.

Vicente-Cheng:
hi @trhoden:

Sorry abut late modified.

I modify the assertion.
I think the test about list return is simple enough so I do not add another conversion for both.

Thanks for your reminder abut unit test assertion :+1: 
If this version is Ok, I will rebase and pull the new request.

vicente

trhoden:
retest this please

trhoden:
Woohoo!  Nice work @Vicente-Cheng.  Go ahead and squash your commits into a smaller set.  

Looks good!

Vicente-Cheng:
hi @trhoden 

Thanks for your help for long time :)
I rebase and pull new request in #271 .

Thanks!!!!
vicente

PR #258, opened 2015-01-09
Enable check_obsoletes in Fedora

trhoden:
Since Fedora ships downstream RPMs that obsolete
python-ceph, those RPMs will be pulled in to resolve
dependencies instead of those hosted by ceph.com, leading
to package errors.

To prevent this, enable check_obsoletes in the Yum
priorities plugin.

Fixes: 10476

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -32,6 +32,24 @@ displayed, similar to::
     are not going to be uninstalled. For example librbd1 and librados which
     qemu-kvm depends on, and removing it would cause issues for qemu-kvm.
 
+Distribution Notes
+------------------
+
+RPMs
+^^^^
+
+On RPM-based distributions, ``yum-plugin-priorities`` is installed to make sure
+that upstream ceph.com repos have a higher priority than distro repos.
+
+Fedora
+^^^^^^
+
+The Fedora distribution maintains packages for Ceph in their own repos, independent
+of the upstream repos on ceph.com.  The packages are structured differently than
+the upstream packages, and can cause dependency resolution errors despite the use
+of the Yum priorities plugin.  To resolve this, ``ceph-deploy`` enables the
+``check_obsoletes`` flag for the Yum priorities plugin.
+
 Specific Releases
 -----------------
 By default the *latest* release is assumed. This value changes when
--- a/ceph_deploy/hosts/fedora/install.py
+++ b/ceph_deploy/hosts/fedora/install.py
@@ -15,6 +15,8 @@ def install(distro, version_kind, version, adjust_repos):
 
     if adjust_repos:
         install_yum_priorities(distro)
+        distro.conn.remote_module.enable_yum_priority_obsoletes()
+        logger.warning('check_obsoletes has been enabled for Yum priorities plugin')
 
         if version_kind != 'dev':
             remoto.process.run(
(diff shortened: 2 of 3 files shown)


trhoden:
cc/ @alfredodeza @ktdreyer

One thing I am struggling with is where to document this behavior so that users understand the difference between installing from upstream vs downstream.  I know this is an area that will change in the coming months as well. I thought about saying something in the README here, but I'm thinking instead that I will submit some doc changes to the ceph install instructions that explain that upstream vs downstream packaging doesn't really play nice together.

If you go with downstream packaging (just yum install from Fedora) you probably need to stick with that.  If you want to use upstream packaging from ceph.com, things will be laid out a bit differently, and you can't really go back and forth between the two.

Thoughts on this?

alfredodeza:
I think it is OK to make a not in the install section: http://ceph.com/ceph-deploy/docs/install.html

And to reference that from ceph.com in the quick install guide.

These changes look good to me for them to be merged, but I will wait for you to confirm extra docs

trhoden:
@alfredodeza I added some notes in install.rst.

Once thing I wanted to check on, though, is that I added a third heading level to the .rst file.  I don't really have a good way to check if the theme/stylesheet on ceph.com handles that okay.  I suspect it does it just fine.  Maybe you know otherwise, though.

I will also take your advice and make sure that the Ceph docs point to the ceph-deploy install notes.  I think there is a lot more details that can be filled in there.

ktdreyer:
It's ok to use 3rd-level headers. I grepped ceph.git and found they're used in the [RBD layering](http://ceph.com/docs/master/dev/rbd-layering/) page.

This looks good to me

`Reviewed-by: Ken Dreyer <kdreyer@redhat.com>`

docs/source/install.rst
```diff
@@ -32,6 +32,24 @@ displayed, similar to::
     are not going to be uninstalled. For example librbd1 and librados which
     qemu-kvm depends on, and removing it would cause issues for qemu-kvm.
 
+Distribution Notes
+------------------
+
+RPMs
+^^^^
+
+On RPM-based distirbutions, yum-plugin-priorities is installed to make sure that
```
alfredodeza: distributions and `yum-plugin-priorities`

docs/source/install.rst
```diff
@@ -32,6 +32,24 @@ displayed, similar to::
...
+upstream ceph.com repos have a higher priority than distro repos.
+
+Fedora
+^^^^^^
+
+The Fedora distribution maintains packages for Ceph in their own repos, independent
+of the upstream repos on ceph.com.  The packages are structured differently than
+the upstream packages, and can cause dependency resolution errors despite the use
+of the Yum priorities plugin.  To resolve this, ``ceph-deploy`` enables the
+check_obsoletes flag for the Yum priorities plugin.
```
alfredodeza: ``check_obsoletes``

ktdreyer:
Depending on whether https://bugzilla.redhat.com/1193182 fixes the issue downstream, we might be able to revert this in ceph-deploy.

ktdreyer:
@branto pointed out in https://bugzilla.redhat.com/1193182 that even if we add version numbers to Fedora and EPEL, it won't fix all cases (eg. Dumpling). So we should just keep this fix in place here.

PR #260, opened 2015-02-16
RHEL packaging split support

trhoden:
This adds a RHEL host module that is meant to deal with the details of installing RHCeph on RHEL machines.  The RHEL module starts from a copy of the centos module, and removes the pieces around Yum priorities and EPEL.

It then always installs the two new packages ceph-mon and ceph-osd.

ceph-deploy will need a bit more work in order to correctly install ceph-mon and ceph-osd only on the needed hosts.  I think that is a bigger revamp that can wait.

The centos host module will still be used on RHEL machines when installing a community version of ceph.  The trigger for whether or not the rhel or centos module is used on a rhel machine is the flag "use_rhceph=true" in the global or install section of cephdeploy.conf.

I tested three scenarios:

A RHEL7 admin and monitor node.  Installed dev version of ceph-deploy on admin node, and enabled "use_rhceph = true" in /root/.cephdeploy.conf after running "ceph-deploy new [mon node]".  On that mon node, I had pre-installed .repo files to point to an internal RH repo that contains the dev versions of RH Ceph. THen, running "ceph-deploy install --no-adjust-repos [mon node]" correctly installed ceph, ceph-mon, and ceph-osd packages and dependencies.  EPEL and yum priorities were not installed.

I cleaned that setup, and tried the same thing without "use_rhceph = true" set in .cephdeploy.conf.  After enabling the server-optional channel, I was able to install the community giant version of Ceph.  So that looks good still.  It only installed the ceph package, thereby using the centos host module, not the RHEL host module.

I then did a CentOS7 setup, and that worked normally as well -- successfully installing the community version as it is currently packaged.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1192579

### DIFF
--- a/ceph_deploy/hosts/__init__.py
+++ b/ceph_deploy/hosts/__init__.py
@@ -6,17 +6,21 @@ on the type of distribution/version we are dealing with.
 """
 import logging
 from ceph_deploy import exc
-from ceph_deploy.hosts import debian, centos, fedora, suse, remotes
+from ceph_deploy.hosts import debian, centos, fedora, suse, remotes, rhel
 from ceph_deploy.connection import get_connection
 
 logger = logging.getLogger()
 
 
-def get(hostname, username=None, fallback=None, detect_sudo=True):
+def get(hostname,
+        username=None,
+        fallback=None,
+        detect_sudo=True,
+        use_rhceph=False):
     """
     Retrieve the module that matches the distribution of a ``hostname``. This
     function will connect to that host and retrieve the distribution
-    informaiton, then return the appropriate module and slap a few attributes
+    information, then return the appropriate module and slap a few attributes
     to that module defining the information it found from the hostname.
 
     For example, if host ``node1.example.com`` is an Ubuntu server, the
@@ -28,6 +32,9 @@ def get(hostname, username=None, fallback=None, detect_sudo=True):
 
     :param hostname: A hostname that is reachable/resolvable over the network
     :param fallback: Optional fallback to use if no supported distro is found
+    :param use_rhceph: Whether or not to install RH Ceph on a RHEL machine or
+                       the community distro.  Changes what host module is
+                       returned for RHEL.
     """
     conn = get_connection(
         hostname,
@@ -48,7 +55,7 @@ def get(hostname, username=None, fallback=None, detect_sudo=True):
             release=release)
 
     machine_type = conn.remote_module.machine_type()
-    module = _get_distro(distro_name)
+    module = _get_distro(distro_name, use_rhceph=use_rhceph)
     module.name = distro_name
     module.normalized_name = _normalized_distro_name(distro_name)
     module.normalized_release = _normalized_release(release)
@@ -62,7 +69,7 @@ def get(hostname, username=None, fallback=None, detect_sudo=True):
     return module
 
 
-def _get_distro(distro, fallback=None):
+def _get_distro(distro, fallback=None, use_rhceph=False):
     if not distro:
         return
 
@@ -77,7 +84,10 @@ def _get_distro(distro, fallback=None):
         'suse': suse,
         }
 
-    return distributions.get(distro) or _get_distro(fallback)
+    if distro == 'redhat' and use_rhceph:
+        return rhel
+    else:
+        return distributions.get(distro) or _get_distro(fallback)
 
 
 def _normalized_distro_name(distro):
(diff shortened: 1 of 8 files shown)


trhoden:
Why thank you ceph-jenkins...  I do need to update some tests.  :)  But I wanted to make sure the direction of this was looking okay first (namely, passing `args` to hosts.get()).

ceph_deploy/hosts/__init__.py
```diff
@@ -6,17 +6,17 @@
...
 from ceph_deploy import exc
-from ceph_deploy.hosts import debian, centos, fedora, suse, remotes
+from ceph_deploy.hosts import debian, centos, fedora, suse, remotes, rhel
 from ceph_deploy.connection import get_connection
 
 logger = logging.getLogger()
 
 
-def get(hostname, username=None, fallback=None, detect_sudo=True):
+def get(hostname, args, fallback=None, detect_sudo=True):
```
GregMeno: @trhoden Is there a reason we are accepting args here instead of expanding the explicit kwargs to include 'use_rhceph' ?

ceph_deploy/hosts/rhel/install.py
```diff
@@ -0,0 +1,94 @@
...
+        distro.conn.remote_module.write_yum_repo(ceph_repo_content)
+
+    if extra_installs:
+        pkg_managers.yum(distro.conn, ['ceph', 'ceph-mon', 'ceph-osd'])
+
+
+def repo_install(distro, reponame, baseurl, gpgkey, **kw):
+    logger = distro.conn.logger
+    # Get some defaults
+    name = kw.pop('name', '%s repo' % reponame)
```
GregMeno: do you mean to mutate kw here?

ceph_deploy/hosts/rhel/mon/create.py
```diff
@@ -0,0 +1,24 @@
...
+        distro.conn,
+        [
+            service,
+            'ceph',
+            '-c',
+            '/etc/ceph/{cluster}.conf'.format(cluster=args.cluster),
+            'start',
+            'mon.{hostname}'.format(hostname=hostname)
+        ],
+        timeout=7,
```
GregMeno: What happens when we timeout?

ceph_deploy/hosts/__init__.py
trhoden: I can definitely do that instead.  The initial reason was that I wanted access to the loaded cephdeploy.conf file, which is stored in args.cd_conf.  But then I didn't actually need that, because when the config file is parsed, all the key/value pairs are adde to args as attributes (first time I've had to use that part of the code).  So I changed the way I was doing it halfway through.

So yes, I could just pass use use_rhceph as  a kwarg instead.

ceph_deploy/hosts/rhel/install.py
trhoden: Hmm, this part was a straight lift from hosts/centos/install.py.

I believe that is something worth looking at.  The mutated `kw` is passed to templates.custom_repo() a bit further down, so it might have some importance.  I'll make a not of it to check out later.

ceph_deploy/hosts/rhel/mon/create.py
trhoden: A `RuntimeError` will be raised by remoto, which gets caught in cephdeploy.mon.mon_create().  That would cause the error to get logged, and an error to the user that it wasn't able to create the monitor.

mon_create() also checks to see that if the service *did* start successfully, that the actual mon status looks normal.

trhoden:
I think this is ready for final review and merge. I updated the main description with final testing results.

@alfredodeza I saw in IRC you had some concerns about the approach being taken here.  Can you elaborate for me?

ceph_deploy/install.py
```diff
@@ -37,7 +37,10 @@ def install(args):
 
     for hostname in args.host:
         LOG.debug('Detecting platform for host %s ...', hostname)
-        distro = hosts.get(hostname, username=args.username)
+        distro = hosts.get(
+            hostname,
+            username=args.username,
+            use_rhceph=bool(getattr(args, 'use_rhceph', False)))
```
alfredodeza: I don't see any flags for `use_rhceph` for CLI consumption, is this set some other way?

trhoden: you are right -- I did not add a CLI switch.  probably should.  The expectation was to have that flag set in cephdeploy.conf by ice-setup, but there should be a switch for it as well.

The annoying thing if it is *not* in cephdeploy.conf is that you also need to do know if we are dealing with RH packages when doing purge and uninstall. I doubt someone wants to pass --use_rhceph if they do ceph-deploy uninstall...  We could just right the flag to the .conf file if we ever see it though, making it persistent.

trhoden:
Closing in favor of pull request #262

PR #261, opened 2015-02-19
Enable check_obsoletes for CentOS and RHEL

trhoden:
We previously enabled this Yum plugin config flag for Fedora, but
recent changes in the EPEL packaging make it necessary for CentOS
and RHEL (when using upstream packages) as well.


trhoden:
based on comments from @ktdreyer on the Fedora implementation, it does seem this is desirable and likely to be permanent. It's pretty straightforward.

I tested it on CentOS -- things looked good.  installs passes when they used to fail.

If I get a simple +1 from anyone on this, I'll merge.

PR #262, opened 2015-02-19
rhel repo support

alfredodeza:
this adds a bit on top of @trhoden work by adding a namespace attribute that tells if a user explicitly asked for a ceph release or not.

This "check for a default and change otherwise" logic will enable us to enforce a behavior where a user in a RHEL system to use the RHEL packages *by default* unless it is specifying otherwise.

Some of the changes are temporary, since this is just to address the immediate need to support split Ceph packages.

### DIFF
--- a/ceph_deploy/install.py
+++ b/ceph_deploy/install.py
@@ -58,7 +58,8 @@ def install(args):
             # upstream. If default_release is True, it means that the user is
             # trying to install on a RHEL machine and should expect to get RHEL
             # packages. Otherwise, it will need to specify either a specific
-            # version, or repo, or a development branch.
+            # version, or repo, or a development branch. Other distro users should
+            # not see any differences.
             use_rhceph=args.default_release,
             )
         LOG.info(
--- a/ceph_deploy/install.py
+++ b/ceph_deploy/install.py
@@ -10,19 +10,33 @@ from ceph_deploy.lib import remoto
 LOG = logging.getLogger(__name__)
 
 
-def install(args):
-    if args.repo:
-        return install_repo(args)
+def sanitize_args(args):
+    """
+    args may need a bunch of logic to set proper defaults that argparse is
+    not well suited for.
+    """
+    if args.release is None:
+        args.release = 'giant'
+        args.default_release = True
 
     # XXX This whole dance is because --stable is getting deprecated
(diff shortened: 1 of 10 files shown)


ceph_deploy/install.py
```diff
@@ -214,7 +237,10 @@ def uninstall(args):
     for hostname in args.host:
         LOG.debug('Detecting platform for host %s ...', hostname)
 
-        distro = hosts.get(hostname, username=args.username)
+        distro = hosts.get(
+            hostname,
+            username=args.username,
+            use_rhceph=bool(getattr(args, 'use_rhceph', False)))
```
trhoden: I think this can be args.default_release now, right?

the args.use_rhceph was intended to be coming from cephdeploy.conf, which doesn't happen anymore.

ceph_deploy/install.py
```diff
@@ -235,7 +261,10 @@ def purge(args):
...
```
trhoden: same here -- args.default_release ?

ceph_deploy/install.py
```diff
@@ -37,7 +51,16 @@ def install(args):
...
-        distro = hosts.get(hostname, username=args.username)
+        distro = hosts.get(
+            hostname,
+            username=args.username,
+            # XXX this should get removed once ceph packages are split for
+            # upstream. If default_release is True, it means that the user is
+            # trying to install on a RHEL machine and should expect to get RHEL
+            # packages. Otherwise, it will need to specify either a specific
+            # version, or repo, or a development branch.
+            use_rhceph=args.default_release,
```
trhoden: To be fair "default release" only means trying to isntall on a RHEL machine if the distro is in fact a RHEL machine.

If the user just did "ceph-deploy install" on an Ubuntu machine, they would get Giant, but the default_release would be true, just would not have an effect, as i understand it.

alfredodeza: correct, I will append an explanation of that case, this might be confusing later on because we are just hijacking the flag to help aid the logic

ceph_deploy/install.py
alfredodeza: I am not sure how I missed these... good catch!

trhoden:
Yes, this looks good to me.  :+1:

PR #264, opened 2015-02-23
Try to remove also ceph-common on Fedora.

mbroz:
Quick workaround but at least it allows purge and purgedata to work
on Fedora 21.

See http://tracker.ceph.com/issues/10928


trhoden:
@mbroz This looks good to me, but can you add your signed-off-by line to the commit?

mbroz:
Ah, ok. Fixed it (but using push -f), I hope review can see the new header with added signed-off. Thanks.

PR #265, opened 2015-03-06
Doc fixes

trhoden:
A doc cleanup, and documenting new behavior in 1.5.22

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,29 +53,33 @@ a contained way.
 
 DEB
 ---
-The DEB repo can be found for ``ceph`` releases in the ``ceph`` repos like::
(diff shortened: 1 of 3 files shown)


alfredodeza:
This looks good although I don't think we should keep maintaining parity between the README and the actual docs.

The plan was to migrate bit by bit towards the official docs and leave links from the README to more explicit sections. But that shouldn't keep this from moving along.

I am ok with whatever you think

trhoden:
@alfredodeza yeah, sounds good.  I'm definitely not adding anything new to the README.  I can slowly start to take things away.  Previously I had fixed the README and not the official docs.

Thanks for looking!

PR #267, opened 2015-03-17
[DNM] Deploy RGW

trhoden:
This is still mostly untested, but wanted to get up for review so it can be poked and get tested early enough to prepare for hammer release.

This is to add "ceph-deploy rgw create..." commands.

### DIFF
--- a/ceph_deploy/gatherkeys.py
+++ b/ceph_deploy/gatherkeys.py
@@ -56,7 +56,7 @@ def gatherkeys(args):
         raise exc.KeyNotFoundError(keyring, args.mon)
 
     # bootstrap
-    for what in ['osd', 'mds']:
+    for what in ['osd', 'mds', 'rgw']:
         keyring = '/var/lib/ceph/bootstrap-{what}/{cluster}.keyring'.format(
             what=what,
             cluster=args.cluster)
--- a/ceph_deploy/gatherkeys.py
+++ b/ceph_deploy/gatherkeys.py
@@ -69,7 +69,11 @@ def gatherkeys(args):
             _hosts=args.mon,
             )
         if not r:
-            raise exc.KeyNotFoundError(keyring, args.mon)
+            if what in ['osd', 'mds']:
+                raise exc.KeyNotFoundError(keyring, args.mon)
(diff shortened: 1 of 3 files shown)


ceph_deploy/gatherkeys.py
```diff
@@ -69,7 +69,11 @@ def gatherkeys(args):
             _hosts=args.mon,
             )
         if not r:
-            raise exc.KeyNotFoundError(keyring, args.mon)
+            if what in ['osd', 'mds']:
+                raise exc.KeyNotFoundError(keyring, args.mon)
+            else:
+                LOG.warning(("No RGW bootstrap key found. Will not be able to "
+                             "deploy RGW daemons"))
```
trhoden: The idea here is just warn if we can't find an RGW bootstrap key.

ceph_deploy/rgw.py
```diff
@@ -0,0 +1,208 @@
...
+    if returncode > 0 and returncode != errno.EACCES:
+        for line in stderr:
+            conn.logger.error(line)
+        for line in stdout:
+            # yes stdout as err because this is an error
+            conn.logger.error(line)
+        conn.logger.error('exit code from command was: %s' % returncode)
+        raise RuntimeError('could not create rgw')
+
+        remoto.process.check(
```
trhoden: I feel like this part can be modified too.  This is a straight copy from the MDS module.  It tries to create "new style" keys, and if that fails falls back to pre-bobtail style keys. Since RGW will only only supported for >= Hammer in this way, I don't see the need for this.  Thoughts?

ceph_deploy/rgw.py
```diff
@@ -0,0 +1,208 @@
...
+            timeout=7
+        )
+    elif init == 'sysvinit':
+        remoto.process.run(
+            conn,
+            [
+                'service',
+                'ceph',
+                'start',
+                'rgw.{name}'.format(name=name),
```
trhoden: Need to double check that the service names are correct.

trhoden:
I also have not added support for modifying the default Civetweb port.

PR #268, opened 2015-03-17
rgw: create works!

liewegas:


trhoden:
Nice, this looks good to me.  :+1: 

I think there's a bit of cleanup I can do later related to the key generation (I don't think we need to support pre-bobtail style keys), but that can be done later.  Looking closer, I think that is already dead code because of the raise().

liewegas:
The two missing pieces now are (1) a teuthology test that uses this, and (2) updated docs describing the streamlined rgw deployment procedure.

trhoden:
@liewegas I think I can work on docs today.  both for ceph-deploy, and perhaps the rgw install section in Ceph.

We'll need to decide when to cut a new version of ceph-deploy, so it can be picked up by a teuthology task.

PR #269, opened 2015-03-19
ceph-deploy doc additions and rework

trhoden:
This make several updates to the docs:

* adds new pages for mds and rgw
* changes the layout of the changelog a bit to make things cleaner/easier to navigate
* fixes some section level mistakes (moving sections up one, down one as far a nesting goes)
* adds release dates in the changelog

All builds and displays correctly for me locally.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -145,7 +145,7 @@ html_static_path = ['_static']
 
 # If true, SmartyPants will be used to convert quotes and dashes to
 # typographically correct entities.
-#html_use_smartypants = True
+html_use_smartypants = False
 
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}
--- a/docs/source/pkg.rst
+++ b/docs/source/pkg.rst
@@ -1,7 +1,7 @@
 
 .. _pkg:
 
-``pkg``
+pkg
 =======
 Provides a simple interface to install or remove packages on a remote host (or
 a number of remote hosts).
@@ -16,7 +16,7 @@ than one package in the argument.
 
 .. _pkg-install:
 
-``--install``
+--install
 -------------
 This flag will use the package (or packages) passed in to perform an installation using
 the distribution package manager in a non-interactive way. Package managers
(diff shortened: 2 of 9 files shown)


docs/source/conf.py
```diff
@@ -145,7 +145,7 @@
 
 # If true, SmartyPants will be used to convert quotes and dashes to
 # typographically correct entities.
-#html_use_smartypants = True
+html_use_smartypants = False
```
alfredodeza: why this change?

trhoden: Some of the reason is in the commit message: https://github.com/trhoden/ceph-deploy/commit/eec4c3555385d76990b7707d005575c1f309405e

But, we have a lot of section headers that start with double-dashes ``--public-network`` for example.  With smarty pants enabled (default) that gets changed to a unicode long dash.  Looks nice, but not good for code.

To work around it, a lost of place used double-back ticks to make it "pre formatted"  My problem with that was that it looked terrible. See: http://ceph.com/ceph-deploy/docs/pkg.html#install

It's dark on top of dark.  :(

So this was my way to get around that.  I wanted to get rid of the backticks to get rid of the dark on dark, but then it turned the double-dashes into long single dashes.  I don't think disable smartypants has any ill-effect (none that I could see).

docs/source/pkg.rst
```diff
@@ -38,7 +38,7 @@ An example call to install a few packages on 2 hosts (with hostnames like
 
 .. _pkg-remove:
 
-``--remove``
+--remove
```
alfredodeza: ah bummer, why remove the back ticks?

trhoden: Is it just my browser where the dark gray on top of light gray looks terrible?

http://ceph.com/ceph-deploy/docs/pkg.html#remove

Any place that is a 2nd level header has a light/medium gray background.  The backticks make it then have an additional box around it that is a darker color.  They just don't seem necessary to me at all.

Here is what's there now:

screen shot 2015-03-19 at 5 24 27 pm

On my local build after the changes, this becomes:

screen shot 2015-03-19 at 5 25 05 pm

I think it is better.  But maybe it is personal taste.

trhoden: Also, removing the backticks isn't just limited to when there are double-dashes.  I think this looks confusing, for example:

screen shot 2015-03-19 at 5 26 53 pm

I just don't think the backticks add anything useful.

trhoden: I could easily pull the backticks and smartypants change (they are intertwined) into a second PR to be debated.  :)

PR #275, opened 2015-04-07
Updating the url used to clone the remoto repository

joehandzik:
Hey guys,

Ran into an issue where the ./bootstrap script used to configure ceph-deploy was trying to pull down a remoto git repo from ceph.com. Updated to direct to Alfredo's remoto repo (heh). I tested this myself, but let me know if there's something I misunderstood. I also took this opportunity to switch to https:// instead of git://...https is much more reliable behind a proxy.

Joe

### DIFF
--- a/vendor.py
+++ b/vendor.py
@@ -64,7 +64,7 @@ def vendor_library(name, version, cmd=None):
             run(['rm', '-rf', vendor_dest])
 
     if not path.exists(vendor_dest):
-        run(['git', 'clone', 'git://ceph.com/%s' % name])
+        run(['git', 'clone', 'https://github.com/alfredodeza/%s.git' % name])
         os.chdir(vendor_src)
         run(['git', 'checkout', version])
         if cmd:
--- a/vendor.py
+++ b/vendor.py
@@ -31,6 +31,8 @@ def run(cmd):
     if result.wait():
         print_error(result.stdout.readlines(), result.stderr.readlines())
 
+    return result.returncode
+
 
 def print_error(stdout, stderr):
     print '*'*80
--- a/vendor.py
+++ b/vendor.py
@@ -64,7 +64,7 @@ def vendor_library(name, version, cmd=None):
             run(['rm', '-rf', vendor_dest])
 
     if not path.exists(vendor_dest):
-        run(['git', 'clone', 'https://github.com/alfredodeza/%s.git' % name])
+        run(['git', 'clone', 'https://github.com/ceph/%s.git' % name])
         os.chdir(vendor_src)
         run(['git', 'checkout', version])
         if cmd:
--- a/vendor.py
+++ b/vendor.py
@@ -66,7 +66,10 @@ def vendor_library(name, version, cmd=None):
             run(['rm', '-rf', vendor_dest])
 
     if not path.exists(vendor_dest):
-        run(['git', 'clone', 'https://github.com/ceph/%s.git' % name])
+        rc = run(['git', 'clone', 'git://ceph.com/%s' % name])
+        if (rc):
+            print "%s: git clone failed using ceph.com url with rc %s, trying github.com" % (path.basename(__file__), rc)
+            run(['git', 'clone', 'https://github.com/ceph/%s.git' % name])
         os.chdir(vendor_src)
         run(['git', 'checkout', version])
         if cmd:
--- a/vendor.py
+++ b/vendor.py
@@ -67,7 +67,7 @@ def vendor_library(name, version, cmd=None):
 
     if not path.exists(vendor_dest):
         rc = run(['git', 'clone', 'git://ceph.com/%s' % name])
-        if (rc):
+        if rc:
             print "%s: git clone failed using ceph.com url with rc %s, trying github.com" % (path.basename(__file__), rc)
             run(['git', 'clone', 'https://github.com/ceph/%s.git' % name])
         os.chdir(vendor_src)


trhoden:
@alfredodeza Which is actually more reliable -- the git mirror on ceph.com, or github?  :)

dmick:
@trhoden good question

joehandzik:
The right answer might be a hybrid of the two strategies then. All I know is that the mirror didn't work for me at all (I assume it's because there's no way to use https instead of git for it, like how github works). My personal vote would be for there to be a fork of  this particular repository in the Ceph project on GitHub, like how many of the other Ceph dependencies are set up in the Ceph ecosystem.

Joe



On Apr 7, 2015, at 5:31 PM, Dan Mick <notifications@github.com<mailto:notifications@github.com>> wrote:


@trhoden<https://github.com/trhoden> good question

�

trhoden:
@joehandzik There actually is a fork already in Ceph's org: https://github.com/ceph/remoto

However, I don't think anything points to it, and it's slightly behind the "upstream" version hosted by @alfredodeza.

I do see your point on git:// vs https:// behind a proxy, though.

dmick:
We should probably just update the github repo.

joehandzik:
My bad, the Alfredo repo came up as the first result and I stuck with that. I can rewrite the patch to point to the Ceph fork instead.

Joe



On Apr 7, 2015, at 8:49 PM, Travis Rhoden <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> There actually is a fork already in Ceph's org: https://github.com/ceph/remoto

However, I don't think anything points to it, and it's slightly behind the "upstream" version hosted by @alfredodeza<https://github.com/alfredodeza>.

I do see your point on git:// vs https:// behind a proxy, though.

�

joehandzik:
Alright, updated the patch. Lemme know if I'm barking up the wrong tree on this one.

ktdreyer:
We really need @alfredodeza 's feedback on this. See https://github.com/alfredodeza/remoto/issues/10 when this came up previously.

dmick:
Why would we need to move the repo?  ceph's copy can continue to be a fork, it just needs to be updated, right?

joehandzik:
Seems reasonable to me...



On Apr 8, 2015, at 12:38 PM, Dan Mick <notifications@github.com<mailto:notifications@github.com>> wrote:


Why would we need to move the repo? ceph's copy can continue to be a fork, it just needs to be updated, right?

�

ktdreyer:
Maybe we should just remove the ceph org fork, and if we *need* patches that Travis or Alfredo are unable to merge (for whatever reason) we can add it back.

Ceph has forked several things; no need to add remoto to that list, right?

dmick:
I...guess?  It seems useful to have such a central Ceph package be more-or-less self-contained in Ceph's github, but I'm not married to it.

This is why I stick to standard easily-available libs.

joehandzik:
From a newbie's perspective, it seemed like forking any github repo that Ceph and ceph utilities depended on was an intentional strategy, bordering on a standard of sorts.

alfredodeza:
A couple of clarifications:

* the ceph organization's fork of `remoto` is a requirement in order for the `ceph.com` Git instance to be able to mirror it.
* The mirror of `remoto` in ceph.com is used for builds, including the build that packages and releases ceph-deploy as we found that to be more reliable than Github.

Having said that, no *user* should be ever be using `./bootstrap` (even worse for a "newbie"). Even in the event of a user trying to develop on ceph-deploy, I would recommend normal Python practices (create a virtualenv, run `python setup.py develop` on it). This "bootstrap" script is mainly there for internal use of our tests as a convenience.

Now, to the specific issue of some url not being available, I suggest not changing the already hard coded url I added in the first place :) Would a fallback be an acceptable compromise?

    try:
        run(['git', 'clone', 'git://ceph.com/%s' % name])
    except Exception as error:
        print "encountered an error using ceph.com --> %s" % error
        run(['git', 'clone', 'https://github.com/ceph/%s.git' % name])

joehandzik:
The README doesn’t say something like, “only use ./bootstrap in certain cases but not others”, so I find that confusing. I’m fine with it being something I shouldn’t touch, but that should be specified in the README a bit more clearly...

While I would definitely classify myself as relatively new to the ceph source, I’m intending that to be a temporary state. :P I ran into this particular issue when I was trying to build RPMs from the github master branches of Ceph and ceph-deploy. The build process calls ./bootstrap.

The fallback you’re proposing is acceptable for my purposes. I can implement in my pull request here, or you could implement in a separate patch to the trunk, both are fine with me!

Joe

joehandzik:
Ok, upated. The change wasn't quite as trivial as planned, the run function was not raising an exception when the git clone failed (which makes sense...'git clone' was being called correctly, it just wasn't working). So I'm passing the return code back to the caller, which works just as well as an exception. 

Do you want two separate pull requests (one for the change to run(), another for the fallback to github change)?

Joe

dmick:
two commits is fine IMO

vendor.py
```diff
@@ -64,7 +66,10 @@ def vendor_library(name, version, cmd=None):
             run(['rm', '-rf', vendor_dest])
 
     if not path.exists(vendor_dest):
-        run(['git', 'clone', 'git://ceph.com/%s' % name])
+        rc = run(['git', 'clone', 'git://ceph.com/%s' % name])
+        if (rc):
```
alfredodeza: no need for parenthesis here

joehandzik: Fair enough, removed in the latest version.

alfredodeza:
OK to test

alfredodeza:
ok to test

PR #278, opened 2015-04-10
fix world readable keyrings

alfredodeza:
Since we couldn't cherry-pick nicely from 3cdc6cb5 due to multiple commits, this PR addresses them separately

### DIFF
--- a/ceph_deploy/new.py
+++ b/ceph_deploy/new.py
@@ -211,18 +211,21 @@ def new_mon_keyring(args):
     keypath = '{name}.mon.keyring'.format(
         name=args.cluster,
         )
-
+    oldmask = os.umask(077)
     LOG.debug('Writing monitor keyring to %s...', keypath)
-    tmp = '%s.tmp' % keypath
-    with file(tmp, 'w') as f:
-        f.write(mon_keyring)
     try:
-        os.rename(tmp, keypath)
-    except OSError as e:
-        if e.errno == errno.EEXIST:
-            raise exc.ClusterExistsError(keypath)
-        else:
-            raise
+        tmp = '%s.tmp' % keypath
+        with file(tmp, 'w') as f:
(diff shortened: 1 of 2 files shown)


trhoden:
Sorry this is seen as a "non-granular" change in master, but it really was pretty straightforward.  It just happened to be on area of code that was changing for other reasons at the same time.

In two files, it's just setting a new umask and saving the old one, then resetting it back after a try/finally block.  4 lines each (8 total).

The diffs look awful because of indenting, and for some other reason I can't figure out why diff has a hard time with it.  :)

There is one other piece of the patch that is missing from this though.  Will add that as a line-comment.

ceph_deploy/new.py
```diff
@@ -211,18 +211,21 @@ def new_mon_keyring(args):
...
-        f.write(mon_keyring)
     try:
-        os.rename(tmp, keypath)
-    except OSError as e:
-        if e.errno == errno.EEXIST:
-            raise exc.ClusterExistsError(keypath)
-        else:
-            raise
+        tmp = '%s.tmp' % keypath
+        with file(tmp, 'w') as f:
```
trhoden: This should now be:

```
        with open(tmp, 'w', 0600) as f:
```

alfredodeza:
What made this a non-granular commit is not the way it looks though, it is basically that we needed just the bits that do the umask and we couldn't apply them because they came with other things in them that we don't want.

I will add the missing line that you spotted.

Again, as granular as possible is preferred always, because it allows us to avoid issues like this one.

ktdreyer:
RH SRT cloned the CVE bug today. I wanted to double-check with you guys that this looks right and it's ok to merge?

trhoden:
I'll give it one more look-over tonight, then merge if looks good. :)
 sorry for delay.

trhoden:
lgtm. merging.

PR #280, opened 2015-04-22
use new remoto version `0.0.25` 

alfredodeza:
That fixes a problem where output would be cut before closing the connection.

Reference issue: http://tracker.ceph.com/issues/11347

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4,6 +4,13 @@ Changelog
 1.5
 ---
 
+1.5.24
+^^^^^^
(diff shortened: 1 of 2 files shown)


docs/source/changelog.rst
```diff
@@ -4,6 +4,13 @@ Changelog
 1.5
 ---
 
+1.5.24
+^^^^^^
+**CHANGEME DATE**
```
trhoden: What about saying **unreleased** instead?

PR #281, opened 2015-05-07
Better handling of custom RGW daemon names

trhoden:
If a custom RGW daemon name is given, prefix it with 'rgw.'.

Update docs and help to clarify this is done.

Fixes: http://tracker.ceph.com/issues/11561

### DIFF
--- a/ceph_deploy/rgw.py
+++ b/ceph_deploy/rgw.py
@@ -176,9 +176,10 @@ def rgw(args):
 
 def colon_separated(s):
     host = s
-    name = 'rgw.' + s
+    name = s
     if s.count(':') == 1:
         (host, name) = s.split(':')
+    name = 'rgw.' + name
     return (host, name)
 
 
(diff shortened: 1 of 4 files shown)


liewegas:
LGTM

ceph_deploy/rgw.py
```diff
@@ -176,9 +176,10 @@ def rgw(args):
 
 def colon_separated(s):
     host = s
-    name = 'rgw.' + s
```
alfredodeza: this needs a test (or two) methinks

trhoden:
@ktdreyer Yeah, I've been told both.  I've been told not to clutter commit messages with links to bugs, and to put them in the PRs instead.  But yeah, I agree.  Unfortunately it does not seem the link between redmine and commit messages is active for ceph-deploy.  If I put "Fixes: <id>" in the commit, it doesn't get pulled into the bug.

ceph_deploy/rgw.py
trhoden: keeping me honest!

trhoden:
@alfredodeza - Tests added!
@ktdreyer - commit messaged updated

ktdreyer:
I'm curious who was advocating for not cluttering commit messages with links to bugs? We should definitely be linking to bugs in commits - it makes cherry-picking and downstream development much saner.

PR #282, opened 2015-05-07
Fix radosgw package name on RPM/EL distros

trhoden:
For RPM based systems, the radosgw package is actually 'ceph-radosgw'.  It's 'radosgw' on Debian.

Make sure we install/uninstall the right package!

### DIFF
--- a/ceph_deploy/hosts/rhel/install.py
+++ b/ceph_deploy/hosts/rhel/install.py
@@ -75,4 +75,4 @@ def repo_install(distro, reponame, baseurl, gpgkey, **kw):
 
     # Some custom repos do not need to install ceph
     if install_ceph:
-        pkg_managers.yum(distro.conn, ['ceph', 'ceph-mon', 'ceph-osd', 'radosgw'])
+        pkg_managers.yum(distro.conn, ['ceph', 'ceph-mon', 'ceph-osd', 'ceph-radosgw'])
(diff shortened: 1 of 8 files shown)


ktdreyer:
@osynge / @dalgaaf ,  we'll also need to adjust `ceph_deploy/hosts/suse/` as well, right?

trhoden:
@ktdreyer I have suse in there already...  since suse points to rpm-hammer, yes I believe it is necessary.

ceph_deploy/hosts/rhel/install.py
```diff
@@ -75,4 +75,4 @@ def repo_install(distro, reponame, baseurl, gpgkey, **kw):
 
     # Some custom repos do not need to install ceph
     if install_ceph:
-        pkg_managers.yum(distro.conn, ['ceph', 'ceph-mon', 'ceph-osd', 'radosgw'])
+        pkg_managers.yum(distro.conn, ['ceph', 'ceph-mon', 'ceph-osd', 'ceph-radosgw'])
```
alfredodeza: this is now going to conflict with my open pull request that addresses the package split

osynge: This would be right for SUSE build -> I will just check but I am fine with either name.

trhoden:
This PR is obsoleted by https://github.com/ceph/ceph-deploy/pull/285.  Closing

PR #283, opened 2015-05-07
Create /var/lib/ceph/radosgw/{cluster}-{name} recursively

trhoden:
Right now the ceph RPM package does not create /var/lib/ceph/radosgw automatically.

ceph-deploy tries to write into  /var/lib/ceph/radosgw/{cluster}-{name}, and fails to create it when /var/lib/ceph/radosgw does not exist. Switch to a os.makedirs call from a os.mkdir call so that it gets done recursively.

This will be safe to keep even if the RPM package starts create /var/lib/ceph/radosgw itself.

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -296,6 +296,17 @@ def safe_mkdir(path):
             raise
 
 
+def safe_makedirs(path):
+    """ create path recursively if it doesn't exist """
+    try:
+        os.makedirs(path)
+    except OSError, e:
+        if e.errno == errno.EEXIST:
+            pass
+        else:
+            raise
+
+
 def zeroing(dev):
     """ zeroing last few blocks of device """
     # this kills the crab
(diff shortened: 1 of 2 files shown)


ktdreyer:
LGTM!

liewegas:
this'll work.. but we can also fix the ceph rpm!

trhoden:
@liewegas I will also submit a patch for the ceph RPM.  Hopefully later tonight, unless someone beats me to it.  But I think this is safe either way.

ktdreyer:
(Note that both RPM and Debian omit this directory, so a patch for ceph.git should include both)

trhoden:
@ktdreyer good point. You know what is really weird, is that I think "ceph-deploy rgw create" works just fine on Debian -- it is only the RPM systems having a problem.  Clearly I need to confirm that.  :/

trhoden:
@ktdreyer - @liewegas pointed out in https://github.com/ceph/ceph/pull/4610 that debian does alread include /var/lib/ceph/radosgw.  He added bootstrap-rgw as well.  So I think we are good on the debian front.  I opened https://github.com/ceph/ceph/pull/4609 to address the RPM side of things.

I still think this safety net in ceph-deploy is a good thing, though.  Will go ahead and merge.

PR #285, opened 2015-05-11
support for package split

alfredodeza:
Only supports the package split by using the new packages for RHEL systems only *unless* specified with the new flags. No other distro is affected although the changes should allow to turn this feature "on" at a later point.

Reference ticket: http://tracker.ceph.com/issues/10894

### DIFF
--- a/ceph_deploy/hosts/__init__.py
+++ b/ceph_deploy/hosts/__init__.py
@@ -61,6 +61,9 @@ def get(hostname,
     module.normalized_release = _normalized_release(release)
     module.distro = module.normalized_name
     module.is_el = module.normalized_name in ['redhat', 'centos', 'fedora', 'scientific']
+    module.is_rpm = module.normalized_name in ['redhat', 'centos',
+                                               'fedora', 'scientific', 'suse']
+    module.is_deb = not module.is_rpm
     module.release = release
     module.codename = codename
     module.conn = conn
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -34,7 +34,9 @@ def repository_url_part(distro):
     return 'el6'
 
 
-def install(distro, version_kind, version, adjust_repos):
+def install(distro, version_kind, version, adjust_repos, **kw):
+    # note: when split packages for ceph land for CentOS, `kw['components']`
+    # will have those. Unused for now.
     logger = distro.conn.logger
     release = distro.release
     machine = distro.machine_type
@@ -124,7 +126,9 @@ def install_epel(distro):
         pkg_managers.yum(distro.conn, 'epel-release')
 
 
-def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True):
+def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True, **kw):
+    # note: when split packages for ceph land for CentOS, `kw['components']`
+    # will have those. Unused for now.
     repo_url = repo_url.strip('/')  # Remove trailing slashes
     gpg_url_path = gpg_url.split('file://')[-1]  # Remove file if present
 
@@ -157,6 +161,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True)
 
 
 def repo_install(distro, reponame, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defauls
+    # note: when split packages for ceph land for CentOS, `packages`
+    # can be used. Unused for now.
+    packages = kw.pop('components', ['ceph', 'ceph-mon', 'ceph-osd'])  # noqa
     logger = distro.conn.logger
     # Get some defaults
     name = kw.pop('name', '%s repo' % reponame)
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -1,7 +1,6 @@
 from ceph_deploy.util import pkg_managers, templates
 from ceph_deploy.lib import remoto
 from ceph_deploy.hosts.util import install_yum_priorities
-from ceph_deploy.util.constants import default_components
 
 
 def rpm_dist(distro):
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -1,6 +1,7 @@
 from ceph_deploy.util import pkg_managers, templates
 from ceph_deploy.lib import remoto
 from ceph_deploy.hosts.util import install_yum_priorities
+from ceph_deploy.util.constants import default_components
 
 
 def rpm_dist(distro):
@@ -165,7 +166,7 @@ def repo_install(distro, reponame, baseurl, gpgkey, **kw):
     # removed them from `kw` so that we don't mess with other defauls
     # note: when split packages for ceph land for CentOS, `packages`
     # can be used. Unused for now.
-    packages = kw.pop('components', ['ceph', 'ceph-mon', 'ceph-osd'])  # noqa
+    packages = kw.pop('components', default_components) or default_components  # noqa
     logger = distro.conn.logger
     # Get some defaults
     name = kw.pop('name', '%s repo' % reponame)
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -163,7 +163,7 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True,
 
 def repo_install(distro, reponame, baseurl, gpgkey, **kw):
     # do we have specific components to install?
-    # removed them from `kw` so that we don't mess with other defauls
+    # removed them from `kw` so that we don't mess with other defaults
     # note: when split packages for ceph land for CentOS, `packages`
     # can be used. Unused for now.
     packages = kw.pop('components', default_components) or default_components  # noqa
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -165,7 +165,7 @@ def repo_install(distro, reponame, baseurl, gpgkey, **kw):
     # removed them from `kw` so that we don't mess with other defaults
     # note: when split packages for ceph land for CentOS, `packages`
     # can be used. Unused for now.
-    packages = kw.pop('components', default_components) or default_components  # noqa
+    packages = kw.pop('components', [])  # noqa
     logger = distro.conn.logger
     # Get some defaults
     name = kw.pop('name', '%s repo' % reponame)
--- a/ceph_deploy/hosts/debian/install.py
+++ b/ceph_deploy/hosts/debian/install.py
@@ -2,6 +2,7 @@ from urlparse import urlparse
 
 from ceph_deploy.lib import remoto
 from ceph_deploy.util import pkg_managers
+from ceph_deploy.util.constants import default_components
 
 
 def install(distro, version_kind, version, adjust_repos, **kw):
@@ -156,7 +157,7 @@ def repo_install(distro, repo_name, baseurl, gpgkey, **kw):
     # removed them from `kw` so that we don't mess with other defauls
     # note: when split packages for ceph land for CentOS, `packages`
     # can be used. Unused for now.
-    packages = kw.pop('components', ['ceph', 'ceph-mon', 'ceph-osd'])  # noqa
+    packages = kw.pop('components', default_components) or default_components
     # Get some defaults
     safe_filename = '%s.list' % repo_name.replace(' ', '-')
     install_ceph = kw.pop('install_ceph', False)
--- a/ceph_deploy/hosts/debian/install.py
+++ b/ceph_deploy/hosts/debian/install.py
@@ -156,7 +156,7 @@ def repo_install(distro, repo_name, baseurl, gpgkey, **kw):
     # removed them from `kw` so that we don't mess with other defaults
     # note: when split packages for ceph land for Debian/Ubuntu, `packages`
     # can be used. Unused for now.
-    packages = kw.pop('components', default_components) or default_components
+    packages = kw.pop('components', [])
     # Get some defaults
     safe_filename = '%s.list' % repo_name.replace(' ', '-')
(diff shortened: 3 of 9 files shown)


ceph_deploy/install.py
```diff
@@ -28,8 +28,41 @@ def sanitize_args(args):
...
+    """
+    flags = {
+        'install_osd': 'ceph-osd',
+        'install_rgw': 'ceph-radosgw',
+        'install_mds': 'ceph-mds',
+        'install_mon': 'ceph-mon',
+        'install_ceph': 'ceph',
+    }
+    if args.install_all:
+        return ['ceph', 'ceph-osd', 'ceph-mds', 'ceph-mon', 'ceph-radosgw']
```
alfredodeza: maybe I should re-use the `default_components` here

osynge:
I have not tested this patch but I like the changes it provides.

ceph_deploy/install.py
```diff
@@ -423,6 +461,41 @@ def make(parser):
...
+        dest='install_mon',
+        action='store_true',
+        help='install install the mon component only',
+    )
+
+    version.add_argument(
+        '--mds',
+        dest='install_mds',
+        action='store_true',
+        help='install install the mds component only',
```
trhoden: s/install install/install

ceph_deploy/install.py
```diff
@@ -423,6 +461,41 @@ def make(parser):
     )
 
     version.add_argument(
+        '--mon',
+        dest='install_mon',
+        action='store_true',
+        help='install install the mon component only',
```
trhoden: s/install install/install

ceph_deploy/install.py
```diff
@@ -423,6 +461,41 @@ def make(parser):
...
+        dest='install_osd',
+        action='store_true',
+        help='install the osd component only',
+    )
+
+    version.add_argument(
+        '--all',
+        dest='install_all',
+        action='store_true',
+        help='install all ceph components (e.g. mon,osd,mds,rgw)',
```
trhoden: Can we mentioned this behavior is the default here?

ceph_deploy/hosts/suse/install.py
```diff
@@ -120,6 +126,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos):
 
 
 def repo_install(distro, reponame, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defauls
+    # note: when split packages for ceph land for CentOS, `packages`
```
trhoden: land for SUSE

ceph_deploy/hosts/rhel/install.py
```diff
@@ -31,10 +35,14 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True)
         distro.conn.remote_module.write_yum_repo(ceph_repo_content)
 
     if extra_installs:
-        pkg_managers.yum(distro.conn, ['ceph', 'ceph-mon', 'ceph-osd'])
+        pkg_managers.yum(distro.conn, packages)
 
 
 def repo_install(distro, reponame, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defauls
```
trhoden: s/defauls/defaults

ceph_deploy/hosts/debian/install.py
```diff
@@ -148,6 +153,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos):
 
 
 def repo_install(distro, repo_name, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defauls
+    # note: when split packages for ceph land for CentOS, `packages`
```
trhoden: s/CentOS/"Debian/Ubuntu"

ceph_deploy/hosts/debian/install.py
```diff
@@ -148,6 +153,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos):
 
 
 def repo_install(distro, repo_name, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defauls
```
trhoden: s/defauls/defaults

ceph_deploy/hosts/centos/install.py
```diff
@@ -157,6 +162,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True)
 
 
 def repo_install(distro, reponame, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defauls
```
trhoden: s/defauls/defaults

ceph_deploy/util/constants.py
```diff
@@ -13,3 +13,12 @@
 mds_path = join(base_path, 'mds')
 
 osd_path = join(base_path, 'osd')
+
+# Default package components to install
+default_components = (
+    'ceph-osd',
+    'ceph-radosgw',
```
trhoden: As you noticed in another thread, this package name is actually different on DEB vs RPM systems.  RPM == ceph-radosgw, and DEB == radosgw.

Even though the package split is not in Debian yet, do we want to handle that now?  Should there be an rpm_default_components and deb_default_components?

Or should there maybe be a get_default_components() method that takes in a host type param, and can then return a different tuple based on the type of host?

trhoden:
Awesome work @alfredodeza!  Mostly just typos in comments, and one question about RPM vs Debian...

As you pointed out in https://github.com/ceph/ceph-deploy/pull/282, this PR and that one will conflict in a not so nice way.

I see two options.  Merge my PR (and the related ones), then do an upstream release to get those bugfixes out.  Then rebase this.

OR

Just let this PR supersede mine, because it also fixes the same problem.

ceph_deploy/install.py
alfredodeza: omg, how did this get here. Will fix

ceph_deploy/install.py
alfredodeza: Absolutely.

trhoden:
@alfredodeza Let me know if this is feature creep, but one other scenario to consider...

We have the "ceph-deploy admin ..." command, that installs the admin keyring to a node.  But it does not install the packages necessary to use the CLI (ceph-common).  It is expected that a user has run "ceph-deploy install" on the node that they want to use as an admin node already.  It would be nice to have a way to do that here as well, and *only* install ceph-common.

ceph-common doesn't show up in the component lists now because everything else depends on it.  It always get's pulled in.  But if you want to create and admin node that only has CLI access, there is no way to do that right now without installing *everything*.  So what about adding a --cli or --common option?

Alternatively (and this could be a different PR), what about just making "ceph-deploy admin" do an install with components="[ceph-common]" automatically?  Would you be opposed to that?

alfredodeza:
@trhoden I have an approach for the packaging difference, I think it would address the needs for this PR and you could just re-use the new `namedtuple` objects in `util.constants` for your PR.

ceph_deploy/hosts/__init__.py
```diff
@@ -61,6 +61,9 @@ def get(hostname,
     module.normalized_release = _normalized_release(release)
     module.distro = module.normalized_name
     module.is_el = module.normalized_name in ['redhat', 'centos', 'fedora', 'scientific']
+    module.is_rpm = module.normalized_name in ['redhat', 'centos',
+                                               'fedora', 'scientific', 'suse']
+    module.is_deb = not module.is_rpm
```
trhoden: This will be handy to have. It does  make me wonder what usecase we have where we are doing something module.is_el != module.is_rpm (meaning RPM, but not SUSE)...  Not important, just something to look at some day.  :)

ceph_deploy/install.py
```diff
@@ -28,8 +28,48 @@ def sanitize_args(args):
...
+    This helper function should parse the args that may contain specifics about
+    these flags and return the default if none are passed in (which is, install
+    everything)
+    """
+    flags = {
+        'install_osd': 'ceph-osd',
+        'install_rgw': 'ceph-radosgw',
+        'install_mds': 'ceph-mds',
+        'install_mon': 'ceph-mon',
+        'install_ceph': 'ceph',
```
trhoden: @ktdreyer I could use your help here...  Do we need "ceph" at all here?  Ceph is a dependency of ceph-osd, ceph-radosgw, ceph-mon, and ceph-mds, right?  In other words, once we do a package split, we never need to in install "ceph".  It's always ceph-*, which will pull in ceph.

trhoden:
@alfredodeza This is looking really good.  I'd like to test it on a few distros (Trusty, Cent 6 and 7, RHEL 7.1) before merging, but I think this might be it!  I'm sure I won't be able to get through all of that until sometime tomorrow though.

Only thing I am wondering is that I _think_ we can remove the "ceph" package from the default components entirely.  post-package split, there is not a use case to "apt-get install ceph" or "yum install ceph".  I pinged @ktdreyer to confirm that one.

ceph_deploy/install.py
ktdreyer: Correct, the "ceph-mon", "ceph-osd", and "ceph-mds" packages will always pull in "ceph". I think of "ceph" as "files common to each of the three servers".

"ceph-radosgw" doesn't depend on "ceph", though, so it can stand alone from "ceph".

trhoden: Cool.  @alfredodeza I think we can nuke this line (and maybe remove it from the description above).

ceph_deploy/util/constants.py
```diff
@@ -13,3 +14,18 @@
 mds_path = join(base_path, 'mds')
 
 osd_path = join(base_path, 'osd')
+
+# Default package components to install
+_base_components = [
+    'ceph-osd',
+    'ceph-mds',
+    'ceph-mon',
+    'ceph',
```
trhoden: And we can remove this one too.  Also update tests, I'm sure.  :)

ceph_deploy/hosts/centos/install.py
```diff
@@ -157,6 +162,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True)
 
 
 def repo_install(distro, reponame, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defaults
+    # note: when split packages for ceph land for CentOS, `packages`
+    # can be used. Unused for now.
+    packages = kw.pop('components', default_components) or default_components  # noqa
```
trhoden: I think these have to be default_components.rpm now?

ceph_deploy/hosts/debian/install.py
```diff
@@ -148,6 +153,11 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos):
 
 
 def repo_install(distro, repo_name, baseurl, gpgkey, **kw):
+    # do we have specific components to install?
+    # removed them from `kw` so that we don't mess with other defaults
+    # note: when split packages for ceph land for Debian/Ubuntu, `packages`
+    # can be used. Unused for now.
+    packages = kw.pop('components', default_components) or default_components
```
trhoden: default_components.deb

ceph_deploy/hosts/rhel/install.py
```diff
@@ -1,13 +1,17 @@
 from ceph_deploy.util import pkg_managers, templates
 from ceph_deploy.lib import remoto
+from ceph_deploy.util.constants import default_components
 
 
-def install(distro, version_kind, version, adjust_repos):
+def install(distro, version_kind, version, adjust_repos, **kw):
+    packages = kw.get('components', default_components) or default_components
```
trhoden: default_components.rpm 

I think this is everywhere.  so I'll just stop here.

ceph_deploy/hosts/centos/install.py
trhoden: @alfredodeza Can you confirm whether this change is necessary?  If it is, I'll wait until it is in then I will pull down this branch and test.

ceph_deploy/install.py
```diff
@@ -28,8 +28,47 @@ def sanitize_args(args):
...
+    """
+    flags = {
+        'install_osd': 'ceph-osd',
+        'install_rgw': 'ceph-radosgw',
+        'install_mds': 'ceph-mds',
+        'install_mon': 'ceph-mon',
+    }
+
+    if distro.is_rpm:
+        defaults = default_components.rpm
```
alfredodeza: @trhoden as you can see here, we are returning either `default_components.rpm` or `default_components.deb`

Instead of changing it in every install method I decided to just do the logic here and adapt. This should (hopefully) address your comments for all the `install.py` files that consume this list, but do let me know if you find this approach bad so we can discuss further :)

ceph_deploy/hosts/centos/install.py
trhoden: Hmm, given your other comment, I'm still slightly confused.  I'm sure you can explain it to me.

kw['components'] is passed in from install.py, and there, components is set by detect_components(), which properly chooses default_components.rpm or .deb, depending on host.  So I fully understand why kw.pop('components') should get the right thing...

But, doesn't kw.pop('components', default_components) assign packages to default_components if kw['components'] does not exist?  And additionally, if it does exist, but evaluates to False, the "or default_components" bit also sets packages = default_components.

And at the top, we have "from ceph_deploy.util.constants import default_components", so isn't that going to just import the NamedTuple?  When using "default_components" directly like this, don't we still have to designing .rpm or .deb?

I may not understand how NamedTuples work.  Will check on that.  :)

alfredodeza:
So a couple of things I just noticed (that I need to fix): 

* the `install --repo` command is meant to install repo files only not packages so we can't always default to install all the packages. We need to do some conflict resolution.

* `install --all` is currently not the default and it should be *unless*  `--repo` is used.

trhoden:
Sweet, I'll give this a shot on RHEL 7.1 to start with.  :)

trhoden:
I tested this on RHEL 7.1 (which has split packages), CentOS 7 (which does not) and Trusty (which does not).  Everything worked!

I saw a few related things we can clean up, but they are all un-related (meaning that existing behavior for non-RHEL has not changed).

Great work @alfredodeza !  :+1:

PR #286, opened 2015-05-12
MDS: systemd enable

goldwynr:
MDS is not enabled if the distribution uses systemd. Add support for distros using systemd.
Also fixed a directory creation fail in case of a new installation while create.

### DIFF
--- a/ceph_deploy/mds.py
+++ b/ceph_deploy/mds.py
@@ -34,7 +34,7 @@ def create_mds(distro, name, cluster, init):
         name=name
         )
 
-    conn.remote_module.safe_mkdir(path)
+    conn.remote_module.makedir(path, [errno.EEXIST])
 
     bootstrap_keyring = '/var/lib/ceph/bootstrap-mds/{cluster}.keyring'.format(
         cluster=cluster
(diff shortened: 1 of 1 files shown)


trhoden:
@goldwynr Can you elaborate on the mkdir piece a bit more for me?

/var/lib/ceph/mds is owned by the "ceph" package, and should be there already whenever ceph is installed.  We should never have to "make" /var/lib/ceph/mds.  However, it may be that this was a recent change?  What version of Ceph did you run into this with?

For example, I recently ran into the exact same thing with RGW.  I fixed that in https://github.com/ceph/ceph-deploy/pull/283.  So it's not that I am opposed to the fix, just want to make sure I understand.

In that referenced PR, a new "safe_mkdirs()" method was introduced.  https://github.com/ceph/ceph-deploy/commit/4bc9e271d5d5d7715dfd36d0142e8aba2911138e

I would use that instead method instead.

ktdreyer:
The RPMs have included the `/var/lib/ceph/mds` directory for a long time (https://github.com/ceph/ceph/commit/1c8ffc5852d573181a29e0cd9b2e83cc46381f71). The first Ceph released version to introduce this directory in a package was v0.59.

goldwynr:
I was using 0.80.8 (without MDS support) and did not realize that the directory was available by the RPM in the sec file. I suppose the systemd fix is still required, right? If yes, I will create a new pull request.

PR #287, opened 2015-05-13
fix `calamari connect`

alfredodeza:
Removes all the dependency on custom repos and relies on previously setup repositories.

Reference ticket: https://bugzilla.redhat.com/show_bug.cgi?id=1219344

### DIFF
--- a/ceph_deploy/calamari.py
+++ b/ceph_deploy/calamari.py
@@ -35,6 +35,8 @@ def connect(args):
             distro.release,
             distro.codename
         )
+        LOG.info('assuming that a repository with Calamari packages is already configured.')
+        LOG.info('Refer to the docs for examples (http://ceph.com/ceph-deploy/docs/conf.html)')
 
         rlogger = logging.getLogger(hostname)
 
@@ -78,7 +80,9 @@ def calamari(args):
 
 def make(parser):
     """
-    Install and configure Calamari nodes
+    Install and configure Calamari nodes. Assumes that a repository with
+    Calamari packages is already configured. Refer to the docs for examples
+    (http://ceph.com/ceph-deploy/docs/conf.html)
     """
     parser.add_argument(
         'subcommand',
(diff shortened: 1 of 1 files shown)


trhoden:
@ktdreyer Can we confirm that the non-RHEL packaging will include the salt-minion pieces in the same repos as mon/osd?

This change will make it such that both RHEL and non-RHEL nodes will install/configure salt-minions by only using the repos they have available for installing ceph-mon and ceph-osd. We know this is the case for RHEL, and we definitely want that to be the case for Ubuntu as well.

Do we know that already?  If not, can we just go ahead and say "yes, we will do that"?  :)

alfredodeza:
We could also mention it when attempting the `connect` command and in the help menu:

    This command assumes that a repository with Calamari packages is already configured. Refer
    to the docs for examples (http://ceph.com/ceph-deploy/docs/calamari.html)

trhoden:
I like that idea, @alfredodeza.  Add that in, and I'm happy to merge!

ktdreyer:
yes, we will do that!

alfredodeza:
With the new changes, the help menu shows:

    $ ceph-deploy calamari --help
    usage: ceph-deploy calamari [-h] [--release [CODENAME]]
                                [--master [MASTER SERVER]]
                                {connect} hosts [hosts ...]
    
    Install and configure Calamari nodes. Assumes that a repository with Calamari
    packages is already configured. Refer to the docs for examples
    (http://ceph.com/ceph-deploy/docs/calamari.html)
    
    positional arguments:
      {connect}
      hosts
    
    optional arguments:
      -h, --help            show this help message and exit
      --release [CODENAME]  Use a given release from repositories defined in ceph-
                            deploy's configuration. Defaults to 'calamari-minion'
      --master [MASTER SERVER]
                            The domain for the Calamari master server

trhoden:
This is looking really good.  From your --help output, and looking at the code, I think that --release is no longer used at all?  That can probably removed entirely. I caught it because it said that it defaulted to 'calamari-minion', which wouldn't be right anymore anways.

trhoden:
Looks good to me now.  Merging!

PR #288, opened 2015-05-14
mds: enable on create with systemd

goldwynr:
Enable MDS services for distros using systemd.

Signed-off-by: Goldwyn Rodrigues <rgoldwyn@suse.com>

### DIFF
--- a/ceph_deploy/mds.py
+++ b/ceph_deploy/mds.py
@@ -108,6 +108,16 @@ def create_mds(distro, name, cluster, init):
             ],
             timeout=7
         )
+    elif init == 'systemd':
+        remoto.process.run(
+            conn,
+            [
+                'systemctl',
+                'enable',
+                'ceph-mds@{name}'.format(name=name),
+            ],
+            timeout=7
+        )
 
     if distro.is_el:
         system.enable_service(distro.conn)


alfredodeza:
OK to test

alfredodeza:
@trhoden I am thinking that this `mds` module needs to be ported to `hosts/**/mds.py` to avoid the if if if elif else dance

alfredodeza:
ok to test

trhoden:
@alfredodeza Yeah, I see what you mean.  Good idea -- make it just like mon. Of course, then rgw should be the same way -- anything with a service. :)

I also thing the whole service start/enable scheme is due for an overhaul.  We still don't properly handle systemd everywhere, really because Ceph itself doesn't do it everywhere yet.  But I think it's time to put it in place a bit better.

liewegas:
@trhoden yeah we're inching toward full systemd support.  this looks right to me... i think we're still missing a ceph-rgw@ service that looks just like these ones

trhoden:
I agree that this looks right...

It will be a blind merge, as I am unable to test this.  Looking through the code-base, the only distro where ceph-deploy attempts to use systemd is openSUSE 13.1.

I grabbed an openSUSE 13.1 vagrant box and tried to use ``ceph-deploy install`` in it, and it failed terribly.  Looking through things, it doesn't look it could be working at all.  I know the SUSE folks have some downstream patches, and this definitely let's me know that a lot of them have to be around getting ceph-deploy to work on SUSE at all.  :)

ceph-deploy tries to point to rpm-{release}/distro for an RPM repo, resulting in things like rpm-hammer/sles11, or rpm-hammer/opensuse12.2.  But ceph.com hasn't hosted a SUSE specific repo since Emperor...

liewegas:
blind is okay for me, all the systemd stuff is a wip at this point.  i'm trying to find time to get my home fedora cluster up so i can test all this in a "real" environment

ktdreyer:
that works for me

PR #289, opened 2015-05-14
don't attempt to install mds in RHEL

alfredodeza:
And display nicely if that is the case

Reference issue: https://bugzilla.redhat.com/show_bug.cgi?id=1218962

### DIFF
--- a/ceph_deploy/mds.py
+++ b/ceph_deploy/mds.py
@@ -128,6 +128,8 @@ def mds_create(args):
 
     bootstrapped = set()
     errors = 0
+    failed_on_rhel = False
+
     for hostname, name in args.mds:
         try:
             distro = hosts.get(hostname, username=args.username)
@@ -138,6 +140,7 @@ def mds_create(args):
                 distro.release,
                 distro.codename
             )
+
             LOG.debug('remote host will use %s', distro.init)
 
             if hostname not in bootstrapped:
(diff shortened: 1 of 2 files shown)


ceph_deploy/mds.py
```diff
@@ -138,6 +138,11 @@ def mds_create(args):
                 distro.release,
                 distro.codename
             )
+            if distro.normalized_name == 'centos':
```
trhoden: This doesn't look right.  It should be ``normalized_name == 'redhat'``, but even that isn't enough.

It's not that we don't want to use MDS on RHEL, it's that we don't want to use MDS on RHEL + RHCS.  If you're using upstream packages, MDS will work just fine.

trhoden: I think this is another argument in favor of moving these things into hosts/\*/{rgw,mod,mds}.py, actually.  If it was setup that way, this could be in the ``hosts/rhel`` module, and *that* only applies to RHEL + RHCS.

alfredodeza: Ungh, it looks like I pushed this by mistake. It was `'redhat'` but I was playing around in CentOS.

The port to modularize this will need to be a separate ticket. It is not hard, but it would be too many changes for this right now.

You raise a good point though and to enforce this rule so that it applies to RHCS it will need to default to not installing it *unless* specified with `--install-mds`. If you agree with that approach I will make the changes and re-push

trhoden: I'm not sure I follow the ``isntall-mds`` logic.  What would this flag apply to, ``ceph-deploy mds create``, or ``ceph-deploy install``?

Truthfully, I feel like no matter which way we approach it, it's going to be a way that effects the default behavior for upstream/community users.  I don't think the change will be backwards compatible.  We have a an idea of how to achieve that, though, which is the modularization you mentioned.  I would suggest waiting for that?

The BZ that brought this up isn't an immediate priority that I can tell.  I think it can wait to be done right.  What are your thoughts?

docs/source/changelog.rst
```diff
@@ -11,6 +11,7 @@ Changelog
 * Use version 0.0.25 of `remoto` that fixes an issue where output would be cut
   (https://github.com/alfredodeza/remoto/issues/15).
 * Automatically prefix custom RGW daemon names with 'rgw.'
+* Refuse to install/deploy MDS in RHEL distros as it is not supported.
```
trhoden: I think this line is not quite true anymore.  It's more that we log a warning that it may not work on RHEL if using RHCS, but there is no refusal to do anything.

PR #290, opened 2015-05-15
Add --cli/--common option to ceph-deploy install

trhoden:
This is a convenience option for users setting "admin" nodes.  If a user is going to run ``ceph-deploy admin`` on a node, the only package it really needs on that node to use the CLI is ``ceph-common``.  We had no way to allow a user to only install that.

This feature builds on top of the package split support, and thus this option only works on distros/versions with the package split in place.  Right now that is only downstream RHCS.

``ceph-common`` does not need to be added the default list of packages anywhere because it is always a dependency when it is needed.

### DIFF
--- a/ceph_deploy/install.py
+++ b/ceph_deploy/install.py
@@ -54,6 +54,7 @@ def detect_components(args, distro):
         'install_rgw': 'ceph-radosgw',
         'install_mds': 'ceph-mds',
         'install_mon': 'ceph-mon',
+        'install_common': 'ceph-common',
     }
 
     if distro.is_rpm:
@@ -503,6 +504,13 @@ def make(parser):
         dest='install_osd',
         action='store_true',
         help='install the osd component only',
+    )
+
+    version.add_argument(
+        '--cli', '--common',
+        dest='install_common',
+        action='store_true',
+        help='install the common component only',
     )
 
     version.add_argument(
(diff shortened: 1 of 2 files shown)


alfredodeza:
@trhoden does this need to be in the defaults as well? or is it pulled in when other split packages are installed (as a dependency) ?

trhoden:
@alfredodeza I think I addressed that in the original PR description.

But no, it does not need to be added anywhere else.  Every package in the list of defaults has ``ceph-common`` as a package dependency.

ktdreyer:
Yep, it's "common" to all of Ceph!! (groan)

ktdreyer:
A couple days ago someone on IRC in #ceph-devel was asking about this PR...

```
< smithfarm_nb> moin https://github.com/ceph/ceph-deploy/pull/290
< smithfarm_nb> this pull request refers to "package split support"
< smithfarm_nb> can someone, loicd maybe?, explain what that is?
```

Note that @trhoden 's branch name is "RM-11642", which is a new convention we're using to indicate "this fixes Redmine issue 11642", ie http://tracker.ceph.com/issues/11642

For broader information about the in-progress Ceph packaging split, see http://tracker.ceph.com/issues/10894 and http://tracker.ceph.com/issues/10587

PR #293, opened 2015-05-20
ceph_deploy/hosts/suse/install.py : SUSE zypper ref

osynge:
zypper may not pull in latest packages if the repository cache is not updated
so we shodul do a 'zypper ref' before installign packages.


osynge:
Im going to close this pull request and make a new one

PR #294, opened 2015-05-21
ceph_deploy/hosts/suse/install.py: SUSE zypper ref

osynge:
zypper may not pull in latest packages if the repository cache is not updated
so we should do a 'zypper ref' before installing packages.


osynge:
this patch has caused ceph testing to break dont know why though

PR #296, opened 2015-05-21
suse.uninstall: ceph-deploy purge fails on SUSE.

osynge:
This is because ceph-deploy tries to remove a package that does not exist.

To fix this we need to change package name to correct name ceph-radosgw
from the old name radosgw.


trhoden:
@osynge The jenkins build bots have been having lots of issues lately.  They were down for a couple of days.  Looks like the tests failed because they either didn't have internet access, or could reach GitHub for another reason.  This is a pretty clear fix.  Thanks!  :+1:

PR #297, opened 2015-05-21
Enable systemd services for suse.

osynge:
Make sure systemd is enabled on the suse platform

### DIFF
--- a/ceph_deploy/mds.py
+++ b/ceph_deploy/mds.py
@@ -122,6 +122,8 @@ def create_mds(distro, name, cluster, init):
     if distro.is_el:
         system.enable_service(distro.conn)
 
+    if distro.distro == 'suse':
+        system.enable_service(distro.conn, 'ceph')
 
 def mds_create(args):
     cfg = conf.ceph.load(args)
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -385,6 +385,8 @@ def activate(args, cfg):
         if distro.is_el:
             system.enable_service(distro.conn)
 
+        if distro.distro == 'suse':
+            system.enable_service(distro.conn, 'ceph')
         distro.conn.exit()
 
 


ktdreyer:
Are you using systemd service unit files in SUSE's downstream packages?

osynge:
@ktdreyer yes we are, as a group we are trying to get everything we have downstream moved into master asap.

osynge:
@ktdreyer I just tested and the files are identical.

trhoden:
@osynge This needs to be rebased against master.

The conflict is surely coming from: https://github.com/ceph/ceph-deploy/commit/ef472ea3403111dff9202674ae4a3013acff0076, which went in a couple of weeks ago.  It also fixed the use of changing services names in util/system.py

As a side note, since this PR is only 4 days old, and the conflicting merge went in 17 days ago, you should have been able to see that the PR could not be merged automatically when creating it.  That's usually an indication of needing to rebase before opening the PR.

That being said, I have seen GitHub screw that up many times.  :)

osynge:
rebased to master.

ceph_deploy/mds.py
```diff
@@ -122,6 +122,8 @@ def create_mds(distro, name, cluster, init):
     if distro.is_el:
         system.enable_service(distro.conn)
 
+    if distro.distro == 'suse':
+        system.enable_service(distro.conn, 'ceph')
```
trhoden: Since ``system.enable_service(distro.conn)`` and ``system.enable_service(distro.conn, 'ceph')`` do the same thing, and since the only different between ``distro.is_el`` and ``distro.is_rpm`` is adding SUSE to ``is_el`` (https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/hosts/__init__.py#L63-L65), can we instead implement this by just chaning ``distro.is_el`` to ``distro.is_rpm``?

I don't think we need an entirely new if case to handle this.

This goes for both this change and the one in osd.py.

osynge: Personally I much prefer the change you propose.

Changing 

distro.is_el -> distro.is_rpm

I would like to see distro.is_el deprecated and removed if at all possible.

trhoden: @osynge point taken!  That seems like a good suggestion.

There are three instances of is_el in use -- two of them are here (osd.py and mds.py), and the other is rgw.py.  If it makes sense to change it for RGW as well (take a look at it) might as well do all three of them.

Thanks!

trhoden:
@osynge I just wanted to revisit this PR.

I think it's just waiting on a refactor to switch from adding new `if` statements to just modifying the existing `is_el` checks to use `is_rpm` instead.

alfredodeza:
This can be re-opened if @osynge can get it rebased and updated

PR #298, opened 2015-05-21
ceph_deploy/mon:ceph-deploy should setup the keys.

osynge:
In my opinion making authorization keys is a
configuration management task not an init script task.

The systemd init scripts in ceph do not make keys
for mon daemon as they do with system V init scripts.


osynge:
Seems this has been fixed by you guys already

PR #299, opened 2015-05-22
Allow alternative cluster names with systemd

osynge:
By modifying the content of:

```
/etc/sysconfig/ceph
```

We can set the cluster name of ceph. This is done via

```
systemd_defaults_clustername
```

This code only executes on systemd systems.

### DIFF
--- a/ceph_deploy/util/system.py
+++ b/ceph_deploy/util/system.py
@@ -1,6 +1,7 @@
 from ceph_deploy.exc import ExecutableNotFound
 from ceph_deploy.lib import remoto
-
+import os.path
+import re
 
 def executable_path(conn, executable):
     """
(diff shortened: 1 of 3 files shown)


osynge:
I suspect other distros want to use this too.

ceph_deploy/util/system.py
```diff
@@ -30,6 +31,52 @@ def is_systemd(conn):
...
+
+    if has_replaced == False:
+        output_lines.append(replacement)
+    for index in range(len(output_lines)-1 , 0,-1):
+        if len(output_lines[index]) == 0:
+            del(output_lines[index])
+        else:
+            break
+    conn.remote_module.write_file(sysconf_file_path, "\n".join(output_lines) + '\n')
+
```
ktdreyer: Mind providing some unit tests for this new `systemd_defaults_clustername` function?

trhoden:
@sage I'm not completely up to date on where things are going with the systemd support.  Is this a feature that can be used by all our system definitions?

osynge:
@trhoden this will work with upstream systemd init files.

ceph_deploy/util/system.py
trhoden: I second @ktdreyer's comment -- it's a lot of file manipulation to have without unit tests.

Also, @alfredodeza may have a recommendation on how to better to do what I think looks like a remote ``sed`` call.

osynge: @ktdreyer 
@trhoden 

Yes unit tests would be good. I will try to remember tomorrow to work on this.

As for replacing this with sed, while the code would be simpler, the end effect, of commenting out the old value and placing the new value on the next lien is too much to do with sed.

@trhoden

osynge: completely forgot about this for too long. Am now making a new branch with a more generic solution that this pr will use to make testing eaiser.

alfredodeza:
custom cluster names are no longer supported by ceph

PR #300, opened 2015-05-22
Fix CVE-2015-4053 world-readable permissions on client.admin key

trhoden:
When using the ``ceph-deploy admin`` command, the client.admin key would get written with 0644 permissions, making it world readable within /etc/ceph.

This PR makes sure that it is written with 0600 permissions, and adds tests for it as well.

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -200,7 +200,11 @@ def write_monitor_keyring(keyring, monitor_keyring):
     write_file(keyring, monitor_keyring)
 
 
-def write_file(path, content):
+def write_file(path, content, directory=None):
+    if directory:
+        if path.startswith("/"):
+            path = path[1:]
+        path = os.path.join(directory, path)
     with file(path, 'w') as f:
         f.write(content)
 
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -200,12 +200,12 @@ def write_monitor_keyring(keyring, monitor_keyring):
     write_file(keyring, monitor_keyring)
 
 
-def write_file(path, content, directory=None):
+def write_file(path, content, mode=0644, directory=None):
     if directory:
         if path.startswith("/"):
             path = path[1:]
         path = os.path.join(directory, path)
-    with file(path, 'w') as f:
+    with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, mode), 'w') as f:
         f.write(content)
 
 
--- a/ceph_deploy/admin.py
+++ b/ceph_deploy/admin.py
@@ -36,7 +36,8 @@ def admin(args):
 
             distro.conn.remote_module.write_file(
                 '/etc/ceph/%s.client.admin.keyring' % args.cluster,
-                keyring
+                keyring,
+                0600,
             )
 
             distro.conn.exit()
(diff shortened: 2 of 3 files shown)


ktdreyer:
Tests look good on my computer (http://fpaste.org/224724/)

trhoden:
Here are tests on mine: http://fpaste.org/224725/43231057/

All pass.

ceph_deploy/hosts/remotes.py
```diff
@@ -200,9 +200,15 @@ def write_monitor_keyring(keyring, monitor_keyring):
...
-def write_file(path, content):
+def write_file(path, content, mode=None, directory=None):
+    if directory:
+        if path.startswith("/"):
+            path = path[1:]
+        path = os.path.join(directory, path)
     with file(path, 'w') as f:
         f.write(content)
+    if mode:
+        os.chmod(path, mode)
```
liewegas: this isn't quite safe--an attacker could open the file after it is created but before it is chowned.  Even chowning before write isn't quite right.   You need to set the mode *at create time* to be totally safe.

trhoden: Thanks Sage.  This thought crossed my mind, and I even looked for the right Python call to do this (open() and file() built-ins don't take a file permissions mode).

I'll have to drop to a lower-level os.open() to do it, I think.  :)

trhoden:
@ktdreyer @sage @dmick 

I think this is ready for re-review.  I changed the way that write_file() works such that the mode is set at file creation time.

I also did some extra tests and verified that the file permissions are reset if they were previously incorrect.  So this way, if someone re-runs ``ceph-deploy admin <node>``, it will make sure that the file perms on the admin keyring are correct.  yay!

I did a quick check of all the callers of write_file() to see if the default mode of 0644 made sense.  It seemed to with one exception.  Right now, all files created by write_file would have been 0644, so the default keeps that backwards compatible.  However, write_file is called by write_monitor_keyring(), and I don't think the monitor keyring should be 0644, it should be 0400.  However, setting that to 0400 in write_monitor_keyring() didn't work on first try. I need to check if the code path is different here between ``mon create-initial`` and ``mon add``.

As you may gather, there is a lot of cleanup/improvement to be made here.  I think if this looks good, though, it is mergeable, and I can address the monitor keyring separately.  I want to see if you think differently though, and would rather see that addressed at the same time as this.

liewegas:
Looks good to me!

dmick:
I'm somewhat surprised that fdopen(open()) is minimal, but I don't have a better answer either.

trhoden:
@dmick Yeah, seriously.  Though it's ``os.fdopen(os.open())``.  Don't want to mix open() with os.open().  But yeah, after looking around, that's the best I could come up with.  os.open() seems to be the only call that allows you to pass a true file mode with file creation.

The downside is that it ties you to POSIX.  So it's not as portable as open().  But, that's kind of the point for both existing...

My references:
http://www.gossamer-threads.com/lists/python/python/990981
http://stackoverflow.com/questions/5624359/write-file-with-specific-permissions-in-python

It definitely works.  But I do not know if there is a more elegant or pythonic way to do it.

dmick:
Yeah, no, in the little bit of research I did it does seem right.  I think os is missing an fopen, personally, but...there you go.

trhoden:
Okay, with a couple of :+1:'s on this thread, I think I'm going to merge.

PR #301, opened 2015-05-25
Some test cleanup

trhoden:
First round of cleanup for tests that I've wanted to do.  Mostly just cleaning up some imports, and renaming some tests to reflect what they actually do.

Added a few new ones for consistency between the CLI commands.

Want to see this run through Tox to see that py26 is okay with my changes.

### DIFF
--- a/ceph_deploy/tests/test_cli_mon.py
+++ b/ceph_deploy/tests/test_cli_mon.py
@@ -1,11 +1,12 @@
 import argparse
 import collections
-import mock
-import pytest
 import subprocess
 
-from ..cli import main
-from .directory import directory
+import pytest
+from mock import Mock, patch, NonCallableMock
+
+from ceph_deploy.cli import _main as main
+from ceph_deploy.tests.directory import directory
 
 
 def test_help(tmpdir, cli):
@@ -45,9 +46,6 @@ def test_bad_no_mon(tmpdir, cli):
     assert err.value.status == 2
 
 
-from mock import Mock, patch
-
-
 def make_fake_connection(platform_information=None):
     get_connection = Mock()
     get_connection.return_value = get_connection
@@ -65,11 +63,11 @@ mon initial members = host1
 """)
 
     ns = argparse.Namespace()
-    ns.pushy = mock.Mock()
-    conn = mock.NonCallableMock(name='PushyClient')
+    ns.pushy = Mock()
+    conn = NonCallableMock(name='PushyClient')
     ns.pushy.return_value = conn
 
-    mock_compiled = collections.defaultdict(mock.Mock)
+    mock_compiled = collections.defaultdict(Mock)
     conn.compile.side_effect = mock_compiled.__getitem__
 
     MON_SECRET = 'AQBWDj5QAP6LHhAAskVBnUkYHJ7eYREmKo5qKA=='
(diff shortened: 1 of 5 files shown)


trhoden:
Not entirely sure what's up with Jenkins here.  I fixed the DNS (for now), which is what caused the first failure.  2nd and 3rd builds were kicked off manually. first one of those hung (don't know why) and I canceled it.  3rd passed just fine.

:shrug:

ceph_deploy/tests/test_cli_mon.py
```diff
@@ -65,11 +63,11 @@ def test_simple(tmpdir, capsys):
 """)
 
     ns = argparse.Namespace()
-    ns.pushy = mock.Mock()
-    conn = mock.NonCallableMock(name='PushyClient')
+    ns.pushy = Mock()
```
alfredodeza: this is a red flag: we are now maintaining tests for libraries that we don't even use anymore (re: pushy)

This test should be removed

ceph_deploy/tests/test_cli_mon.py
```diff
@@ -81,9 +79,9 @@ def _create_mon(cluster, get_monitor_secret):
     fake_ip_addresses = lambda x: ['10.0.0.1']
     try:
         with patch('ceph_deploy.new.net.ip_addresses', fake_ip_addresses):
-            with mock.patch('ceph_deploy.new.net.get_nonlocal_ip', lambda x: '10.0.0.1'):
-                with mock.patch('ceph_deploy.new.arg_validators.Hostname', lambda: lambda x: x):
-                    with mock.patch('ceph_deploy.new.hosts'):
+            with patch('ceph_deploy.new.net.get_nonlocal_ip', lambda x: '10.0.0.1'):
```
alfredodeza: another reason for removal: this nesting is insane. 

Of course I am partly to blame, but these were adapted when poor ceph-deploy didn't have much of tests anyway. The cost of maintenance is far larger than the benefits (for this case and all others that have multiple nested patching)

ceph_deploy/tests/test_cli_mon.py
trhoden: I completely agree.  That's part of my goal for test cleanup here, and one reason why I called it "round one".  haha.  Just wanted to get things into a consistent place first.

Interestingly, I did play with this exact test a bit first, and on first try the things I tried to remove related to mocking the push stuff made things fail (in ways I though weird, given that it doesn't exist).  But I decided I would come back to it later.

ceph_deploy/tests/test_cli_new.py
```diff
@@ -19,6 +21,18 @@ def test_help(tmpdir, cli):
...
 
 
+def test_bad_no_mon(tmpdir, cli):
+    with pytest.raises(cli.Failed) as err:
+        with cli(
+            args=['ceph-deploy', 'new'],
+            stderr=subprocess.PIPE,
+            ) as p:
+            result = p.stderr.read()
+    assert 'usage: ceph-deploy new' in result
```
alfredodeza: multiple assertions in unit tests are not ideal because when it fails, `bad_no_mon` will not map to the exit status or 'too few arguments' or 'usage:'.

If there was a unit test for each (or for all the cases you deem useful) then it would be obvious and succinct:

    TestMonSubcommand::test_not_enough_arguments
    TestMonSubcommand::test_non_zero_exit_status
    TestMonSubcommand::test_show_usage_example

Would make it easy to tell what is going on and to keep the granularity

trhoden: @alfredodeza yeah, good point.  An excellent area to clean up.  I took the multiple asserts directly from the existing tests for ``--help``, e.g. https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/tests/test_cli_install.py#L11-L13

trhoden:
@alfredodeza One general question -- which of your comments would you like to see addressed before being okay with merging these changes?  I think this will be a long series of changes, and I agree with all of your comments. I just want to bite it off into manageable chunks.  :)

alfredodeza:
@trhoden nothing that you changed here should prevent you from moving forward. Sorry, I should've made it a bit more clear that nothing needed fixing. 
However, I do think that we should just nix the deeply nested tests. Don't fix them. Remove them. If you feel that is terribly cumbersome at the moment, I am OK to revise later.

Very good work here. Thanks so much for the effort!

trhoden:
I have some work queued up that will also modify tests, so I'd like to merge this and rebase on top of it.

PR #302, opened 2015-05-25
[BZ-1221830] radosgw pkg is ceph-radosgw on RPM systems

trhoden:
I incorrectly thought in https://github.com/ceph/ceph-deploy/pull/282 that these changes were no longer needed.

### DIFF
--- a/ceph_deploy/hosts/rhel/uninstall.py
+++ b/ceph_deploy/hosts/rhel/uninstall.py
@@ -7,7 +7,7 @@ def uninstall(conn, purge=False):
         'ceph-common',
         'ceph-mon',
         'ceph-osd',
-        'radosgw'
+        'ceph-radosgw'
         ]
 
     pkg_managers.yum_remove(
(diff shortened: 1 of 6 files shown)


ceph_deploy/hosts/rhel/uninstall.py
```diff
@@ -7,7 +7,7 @@ def uninstall(conn, purge=False):
         'ceph-common',
         'ceph-mon',
         'ceph-osd',
-        'radosgw'
+        'ceph-radosgw'
```
alfredodeza: regardless if this wasn't installed, we are always removing it?

ktdreyer: @alfredodeza that seems ok to me... does it matter?

trhoden: that's what we've been doing for a while.  yum just throws a warnings, but marches on.  I *do* think that a later enhancement we can make, post package-split everywhere, is uninstalling specific packages as well.  Though we should check to see if there is a real usecase for that.

PR #303, opened 2015-05-27
allow to skip rpm imports with `cephdeploy.conf`

alfredodeza:
Setting a default flag in `cephdeploy.conf` flag should work, except that for this specific case, where a boolean flag was needed (vs. a flag with a value) it was not getting parsed correctly.

This PR addresses this problem and documents the example that was needed for the related issue.

Reference ticket: https://bugzilla.redhat.com/show_bug.cgi?id=1219310

### DIFF
--- a/ceph_deploy/conf/cephdeploy.py
+++ b/ceph_deploy/conf/cephdeploy.py
@@ -101,6 +101,7 @@ def set_overrides(args, _conf=None):
     subcommand = args.func.__name__
     command_section = 'ceph-deploy-%s' % subcommand
     conf = _conf or load()
+
     for section_name in conf.sections():
         if section_name in ['ceph-deploy-global', command_section]:
             override_subcommand(
@@ -121,8 +122,22 @@ def override_subcommand(section_name, section_items, args):
     """
     # XXX We are not coercing here any int-like values, so if ArgParse
     # does that in the CLI we are totally non-compliant with that expectation
+    # but we will try and infer a few boolean values
+
+    # acceptable boolean states for flags
+    _boolean_states = {'yes': True, 'true': True, 'on': True,
+                       'no': False, 'false': False, 'off': False}
+
     for k, v, in section_items:
-        setattr(args, k, v)
+        # get the lower case value of `v`, fallback to the booleanized
+        # (original) value of `v`
+        try:
+            normalized_value = v.lower()
+        except AttributeError:
+            # probably not a string object that has .lower
+            normalized_value = v
+        value = _boolean_states.get(normalized_value, v)
+        setattr(args, k, value)
     return args
 
 
--- a/ceph_deploy/tests/unit/test_conf.py
+++ b/ceph_deploy/tests/unit/test_conf.py
@@ -1,5 +1,6 @@
 from cStringIO import StringIO
 from textwrap import dedent
+import pytest
 from mock import Mock, patch
 from ceph_deploy import conf
 from ceph_deploy.tests import fakes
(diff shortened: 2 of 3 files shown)


ktdreyer:
very cool, thanks!

ceph_deploy/conf/cephdeploy.py
```diff
@@ -121,8 +122,23 @@ def override_subcommand(section_name, section_items, args):
     """
     # XXX We are not coercing here any int-like values, so if ArgParse
     # does that in the CLI we are totally non-compliant with that expectation
+    # but we will try and inferr a few boolean values
```
trhoden: s/inferr/infer

ceph_deploy/tests/unit/test_conf.py
```diff
@@ -167,3 +172,21 @@ def test_override_foo_section(self):
         self.conf.items = Mock(return_value=(('bar', 1),))
         arg_obj = conf.cephdeploy.set_overrides(self.args, self.conf)
         assert arg_obj.bar == 1
+
+    @pytest.mark.parametrize('value', truthy_values)
+    def test_override_truthy_values(self, value):
+        self.conf.sections = Mock(
+            return_value=['ceph-deploy-global', 'ceph-deploy-install']
+        )
+        self.conf.items = Mock(return_value=(('bar', 'on'),))
```
trhoden: s/'on'/value

alfredodeza: just fixed! thanks for catching that

trhoden:
@alfredodeza Cool - this seems like a much-needed addition. Thanks!

I think a big part of this ticket is unclear how to address.  It certainly has a packaging component to it.  Basically, we would want someone who yum or apt-get install's ceph-deploy to get a default config file out of the box.  That default file needs to be a separate file in git, so that we can include it in the package, but also so that downstream packages can patch it.

Right now, ceph-deploy automatically writes out a .cephdeploy.conf file in a users $HOME if one is not found, based on the template in https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/conf/cephdeploy.py.  It looks for a config file in the CWD, then in $HOME.

I know there are a lot of historical reasons for this.  I think we might need to take a different approach, however.

If we ship a default config file in a package, it really should be at a place like /etc/cephdeploy/cephdeploy.conf, and owned by the package.  We could then allow users to *override* things in the main file with a .cephdeploy.conf file in their ~, but they need to be additive, not one replacing the other (IMO).

I think looking for a config file in the CWD is problematic, but such a change would have implications.  I kind of felt a config file revamp like this might warrant a 1.6.0 release, because it would break some backwards compatibility.

To wrap up, I think we want to take a lot of the info that is in code right now and move into the config file to shipped with packages.  What I am not sure, however, is how we approach users who install through PyPI.  If not installing with a package, what is the best way to make sure the user has a valid config file?  default values in the code? (seems at odds with the goals of this, but also makes the most sense) Having ceph-deploy maintain writing out a new .cephdeploy.conf to $HOME if it doesn't find one? (Doesn't seem much different than default values actually).

I'm not sure what is the normal practice here when packages and PyPI may compete, and PyPI installs may not include all the same files/locations as an OS package.  I suppose maybe there is a third location where a file could exist, such that PyPI installs can use it (and it works in a virtualenv?).  I'm just talking out loud with the last parts here...

alfredodeza:
I think we can discuss and address the issues you note separately from these changes, as we are only allowing something that should've worked (it didn't) to actually work correctly.

Regardless if we end up providing a way for a packager to include (or not) configuration files. We will still need these changes.

trhoden:
@alfredodeza Fully agreed.  Just wanted to re-iterate the long term goal. Perhaps the wrong forum.  :)  This looks good to me once the small typo is fixed.

trhoden:
:+1:

PR #304, opened 2015-05-28
at runtime, display all flags and values

alfredodeza:
Example output:

    (ceph-deploy)papaya-2 ~/tmp/foo ᓆ ceph-deploy new node2
    [ceph_deploy.conf][DEBUG ] found configuration file at: /Users/alfredo/tmp/foo/cephdeploy.conf
    [ceph_deploy.cli][INFO  ] Invoked (1.5.25): /Users/alfredo/.virtualenvs/ceph-deploy/bin/ceph-deploy new node2
    [ceph_deploy.cli][INFO  ] ceph-deploy options:
    [ceph_deploy.cli][INFO  ]  username                      : None
    [ceph_deploy.cli][INFO  ]  func                          : <function new at 0x103e23aa0>
    [ceph_deploy.cli][INFO  ]  verbose                       : True
    [ceph_deploy.cli][INFO  ]  overwrite_conf                : False
    [ceph_deploy.cli][INFO  ]  prog                          : ceph-deploy
    [ceph_deploy.cli][INFO  ]  quiet                         : False
    [ceph_deploy.cli][INFO  ]  cd_conf                       : <ceph_deploy.conf.cephdeploy.Conf instance at 0x103e46098>
    [ceph_deploy.cli][INFO  ]  cluster                       : ceph
    [ceph_deploy.cli][INFO  ]  ssh_copykey                   : True
    [ceph_deploy.cli][INFO  ]  mon                           : ['node2']
    [ceph_deploy.cli][INFO  ]  public_network                : None
    [ceph_deploy.cli][INFO  ]  ceph_conf                     : None
    [ceph_deploy.cli][INFO  ]  bar                           : '1'
    [ceph_deploy.cli][INFO  ]  cluster_network               : None
    [ceph_deploy.cli][INFO  ]  default_release               : False
    [ceph_deploy.cli][INFO  ]  fsid                          : None
    [ceph_deploy.new][DEBUG ] Creating new cluster named ceph
    [ceph_deploy.new][INFO  ] making sure passwordless SSH succeeds
    [node2][DEBUG ] connected to host: papaya-2.local
    [node2][INFO  ] Running command: ssh -CT -o BatchMode=yes node2

Reference issue: http://tracker.ceph.com/issues/11788

### DIFF
--- a/ceph_deploy/tests/unit/test_cli.py
+++ b/ceph_deploy/tests/unit/test_cli.py
@@ -1,0 +1,46 @@
+from ceph_deploy import cli
+from ceph_deploy.tests import util
+
+
+class FakeLogger(object):
+
+    def __init__(self):
+        self._calls = []
+        self._info = []
+
+    def _output(self):
+        return '\n'.join(self._calls)
+
+    def _record(self, level, message):
+        self._calls.append(message)
+        method = getattr(self, '_%s' % level)
+        method.append(message)
+
+    def info(self, message):
+        self._record('info', message)
+
+
+class TestLogFlags(object):
+
+    def setup(self):
+        self.logger = FakeLogger()
+
+    def test_logs_multiple_object_attributes(self):
+        args = util.Empty(verbose=True, adjust_repos=False)
+        cli.log_flags(args, logger=self.logger)
+        result = self.logger._output()
+        assert ' verbose ' in result
+        assert ' adjust_repos ' in result
+
+    def test_attributes_are_logged_with_values(self):
+        args = util.Empty(verbose=True)
+        cli.log_flags(args, logger=self.logger)
+        result = self.logger._output()
+        assert ' verbose ' in result
+        assert ' : True' in result
+
+    def test_private_attributes_are_not_logged(self):
+        args = util.Empty(verbose=True, _private='some value')
+        cli.log_flags(args, logger=self.logger)
+        result = self.logger._output()
+        assert ' _private ' not in result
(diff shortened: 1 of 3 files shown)


trhoden:
Very helpful indeed!

ceph_deploy/tests/unit/test_cli.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def setup(self):
+        self.logger = FakeLogger()
+
+    def test_logs_multiple_object_attributes(self):
+        args = util.Empty(verbose=True, adjust_repos=False)
+        cli.log_flags(args, logger=self.logger)
+        result = self.logger._output()
+        assert ' verbose ' in result
+        assert ' adjust_repos ' in result
```
trhoden: Since we were just discussing it the other day -- when is it okay to use multiple asserts like this?  :)

alfredodeza: very good question. I felt this was OK here because the test wants to ensure that "not only one value is sent out". 

Since the signature specifies this one assert failure (even though there are two) would indicate that not all of them were present and clearly point to the failure.

This could've been a single assert but the string output was just too big to compare them. It was easier to do two.

Usually, I split them when the answer to the question is different from what the name of the test defines. For this test, I would ask: "is this assert asserting that multiple object attributes are logged?". If I had another assert, maybe something like `assert logger.info.called_once()` The answer to my question would no longer hold.

PR #305, opened 2015-05-28
Use git.ceph.com for GPG keys

trhoden:
Grab the GPG release/autobuild key from git.ceph.com instead of ceph.com/git.

Do some cleanup to make this URL only be defined in one place.

Ref: http://tracker.ceph.com/issues/11406

### DIFF
--- a/ceph_deploy/conf/cephdeploy.py
+++ b/ceph_deploy/conf/cephdeploy.py
@@ -37,7 +37,7 @@ cd_conf_template = """
 # enabled=1
 # gpgcheck=1
 # type=rpm-md
-# gpgkey=https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/autobuild.asc
+# gpgkey=https://git.ceph.com/?p=ceph.git;a=blob_plain;f=keys/release.asc
 
 # apt repos:
 # [myrepo]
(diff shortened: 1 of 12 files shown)


ceph_deploy/util/paths/gpg.py
```diff
@@ -0,0 +1,8 @@
+from ceph_deploy.util import constants
+
+def url(key_type, protocol="https"):
+    return "{protocol}://{url}{key_type}.asc".format(
+        protocol=protocol,
+        url=constants.gpg_key_base_url,
```
alfredodeza: yessssssssss

ceph_deploy/conf/cephdeploy.py
```diff
@@ -37,7 +37,7 @@
 # enabled=1
 # gpgcheck=1
 # type=rpm-md
-# gpgkey=https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/autobuild.asc
+# gpgkey=https://git.ceph.com/?p=ceph.git;a=blob_plain;f=keys/release.asc
```
alfredodeza: can't we just re-use the util/path thing here?

trhoden: Ack - yes we can.  For some reason I had this in my head as "docs", like the .rst files. Silly of me...  :) Nice catch!

trhoden:
@alfredodeza added 1 commit to take care of the url in the cephdeploy.conf template.

PR #306, opened 2015-05-28
[RM-11572] use http to get GPG key on Wheezy

trhoden:
For whatever reason, Debian Wheezy will not accept the SSL cert on git.ceph.com, though other distros seem to be fine with it.  So, switch to http for that distro.

Reference: http://tracker.ceph.com/issues/11572


alfredodeza:
At some point I think we should leverage configuration over code changes. I know we've talked about this, but lets get that on a ticket at some point so we can track that.

PR #307, opened 2015-06-02
support for distro Mint which is based on ubuntu

honkiko:
Signed-off-by: Hong Zhiguo <zhiguohong@tencent.com>

### DIFF
--- a/ceph_deploy/hosts/__init__.py
+++ b/ceph_deploy/hosts/__init__.py
@@ -103,6 +103,8 @@ def _normalized_distro_name(distro):
         return 'suse'
     elif distro.startswith('centos'):
         return 'centos'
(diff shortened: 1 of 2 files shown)


honkiko:
Works on my "Linux Mint 17 Qiana". Should also works on other recent versions.

alfredodeza:
ok to test

alfredodeza:
@honkiko I don't see any issues with this change, but Mint is not one of the officially supported distributions. Having it here in ceph-deploy would imply that we are supporting it.

ktdreyer:
Personally I don't see the harm in taking this - "supported" is a fuzzy word :)

liewegas:
Yeah, I think we should make ceph-deploy work with as many target distros as is practical.

PR #308, opened 2015-06-12
Fix package installation on SUSE platforms

ddiss:
As raised:


osynge:
I suggest this is nicer than my kill all the code in this area solution :)

ddiss:
Submitted with unit test via https://github.com/ceph/ceph-deploy/pull/312 .

PR #310, opened 2015-06-16
[BZ-1231203] throws if multiple mon are passed to "mon add"

tchaikov:
Signed-off-by: Kefu Chai <tchaikov@gmail.com>

### DIFF
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -175,6 +175,8 @@ def mon_add(args):
 
     if not args.mon:
         raise exc.NeedHostError()
+    elif len(args.mon) > 1:
+        raise exc.GenericError('Only one node can be added at a time')
     mon_host = args.mon[0]
 
     try:
@@ -482,7 +484,8 @@ def make(parser):
 
       If the section for the monitor exists and defines a `mon addr` that
       will be used, otherwise it will fallback by resolving the hostname to an
-      IP. If `--address` is used it will override all other options.
+      IP. If `--address` is used it will override all other options. Please
+      note that only one node can be added at a time.
 
     destroy
       Completely remove monitors on a remote host. Requires hostname(s) as
--- a/docs/source/mon.rst
+++ b/docs/source/mon.rst
@@ -73,7 +73,7 @@ destroy
 Completely remove monitors on a remote host. Requires hostname(s) as
 arguments::
 
-    ceph-deploy mon destroy node1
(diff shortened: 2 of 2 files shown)


tchaikov:
we have py26 in tox.ini
```ini
[tox]
envlist = py26, py27, flake8
```

but seems we don't have python2.6 on the jenkins slave:
```
ERROR:   py26: InterpreterNotFound: python2.6
  py27: commands succeeded
  flake8: commands succeeded
```

tchaikov:
@alfredodeza shall we install python2.6 on the slaves? or we should change the build matrix for this test job in jenkins?

alfredodeza:
@tchaikov good question, I think we are OK for now, let me review this first :)

alfredodeza:
@tchaikov this looks good. Would you mind adding a note in the help menu and the docs? That way we are even more explicit about this.

tchaikov:
sure thing!

ceph_deploy/mon.py
```diff
@@ -482,7 +484,8 @@ def make(parser):
 
       If the section for the monitor exists and defines a `mon addr` that
       will be used, otherwise it will fallback by resolving the hostname to an
-      IP. If `--address` is used it will override all other options.
+      IP. If `--address` is used it will override all other options. Please
+      note that only one node can be added at a time.
```
alfredodeza: this is good. Now the only one left is in ceph-deploy/docs/sources/mon.rst

tchaikov:
@alfredodeza thanks for your review! updated & repushed,

trhoden:
As a note to the future, seems like a better improvement would be to remove `nargs='*'`.  No need for `nargs` at all, since `mon` is required, and we only want one host listed.  with `'*'`, it will still allow an empty list, which is not what we want.  `'+'` would at least require 1, but the check to seen that `len(args.mon) > 1` will become totally unnecessary.

trhoden:
and if we require a list for `args.mon`, could just change to `nargs=1`.  But I see no reason for the list.

tchaikov:
> we only want one host listed.

@trhoden probably we need to use `add_parser()` instead, because all the subcommands are sharing the same `args.mon`. for example, both `mon create` and `mon destroy` accept multiple mons. but this does not apply to `mon add`.

alfredodeza:
Right, we are using `args.mon` in several places, all with different requirements.

The one other approach would be to make `mon_add` support multiple monitors, but that seemed like a lot more work

PR #311, opened 2015-06-23
Add tests for argparse

trhoden:
Throwing this up for early review.  Adding tests to validate that we are getting the desired behavior from our argument parsing.

We already had a parser factory, which means it was already set up correctly for testing.

In addition to new test coverage, the advantage here is that we can start to get rid of some tests that were running the program (i.e. invoking ceph-deploy the command) in order to test the arg parsing, which is totally unnecessary and also takes longer.  This removes complexity by not having to use the subprocess module, and should ultimately result in faster tests, too.


trhoden:
For anyone following along, most of these jenkins errors are because the build slaves no longer have Python 2.6.

@alfredodeza Do we plan to add that back in?  If we don't, we should remove it from tox since all build checks will fail.

alfredodeza:
@trhoden I think we should keep it in tox and fix the slaves, but fixing the slaves is not straight forward at the moment :(

alfredodeza:
YESSSSSSSSSSSS

ktdreyer:
yes

trhoden:
Okay, @alfredodeza, @dmick, @ktdreyer this is ready for review.

It seems big, but it's just adding a bunch of test with similar structure.  One file for each "ceph-deploy {cmd}", plus a top-level one.

There are a bunch of tests that are marked skip for now, all referencing one of four issues:

http://tracker.ceph.com/issues/12147
http://tracker.ceph.com/issues/12150
http://tracker.ceph.com/issues/12151
http://tracker.ceph.com/issues/12168

I did not create a file for "pkg" just yet, as I would redo the whole thing after http://tracker.ceph.com/issues/12171 anyways.

The idea here is test the argument parsing without actually invoking the program.  Previously we had to invoke "ceph-deploy" in the virtualenv to see if we hit any parsing problems.  I want to have this baseline to look for regressions because I want to make some changes to the argparse config to create much better help menus.  There are way too many places where options are allowed that do nothing for the given command.

Also, I don't know why the Jenkins tests are failing right now.  All the tests pass, and at the end it show:
``` 
  py26: commands succeeded
  py27: commands succeeded
  flake8: commands succeeded
  congratulations :)
ERROR: Publisher com.cloudbees.jenkins.GitHubCommitNotifier aborted due to exception
java.io.IOException: Cannot determine sha1 of the commit. The status cannot be reported
```


Here are all the new tests:
```
travis$ py.test parser
============================================ test session starts =============================================
platform darwin -- Python 2.7.6 -- py-1.4.27 -- pytest-2.7.0
rootdir: /Users/travis/github/ceph-deploy-trhoden, inifile: setup.cfg
collected 226 items

parser/test_admin.py ....
parser/test_calamari.py ......
parser/test_config.py ....s..s..
parser/test_disk.py ......s...ss...ss......s...s...
parser/test_gatherkeys.py ....
parser/test_install.py ......s....s...............s..s........
parser/test_main.py .................................
parser/test_mds.py .s..
parser/test_mon.py .ss...s..ss....s..s.sss..
parser/test_new.py .....s........
parser/test_osd.py ......s...s...s...ss...ss...ss......s...
parser/test_purge.py ....
parser/test_purgedata.py ....
parser/test_rgw.py .s..
parser/test_uninstall.py ....

=================================== 190 passed, 36 skipped in 2.36 seconds ===================================
(travis)sombrero:tests travis$
```

ceph_deploy/tests/parser/test_install.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def test_install_multiple_hosts(self):
+        hostnames = ['host1', 'host2', 'host3']
+        args = self.parser.parse_args(['install'] + hostnames)
+        assert frozenset(args.host) == frozenset(hostnames)
+
+    def test_install_release_default_is_none(self):
+        args = self.parser.parse_args('install host1'.split())
+        assert args.release is None
+        assert args.version_kind == "stable"
```
alfredodeza: this is a good reason to use two asserts in a test

ceph_deploy/tests/parser/test_install.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def test_install_adjust_repos_false(self):
+        args = self.parser.parse_args('install --no-adjust-repos host1'.split())
+        assert not args.adjust_repos
+
+    @pytest.mark.skipif(reason="http://tracker.ceph.com/issues/12147")
+    def test_install_adjust_repos_false_with_custom_release(self):
+        args = self.parser.parse_args('install --release firefly --no-adjust-repos host1'.split())
+        assert args.release == "firefly"
+        assert not args.adjust_repos
```
alfredodeza: why `assert not` instead of asserting the positive ? (not sure what args.adjust_repos is, False?)

trhoden: I just was going off the fact that adjust_repos has to be True of False.  I can change to ```assert args.adjust_repos is False``` if that is more clear.

PR #312, opened 2015-06-29
Fix package installation on SUSE platforms v2

ddiss:
As raised:

Changes since v1 (https://github.com/ceph/ceph-deploy/pull/308):
- Test ceph-deploy component -> SUSE package name mapping (Thanks Ken)
- Rebase against latest upstream

### DIFF
--- a/ceph_deploy/hosts/suse/install.py
+++ b/ceph_deploy/hosts/suse/install.py
@@ -24,16 +24,15 @@ def map_components(components):
 
 
 def install(distro, version_kind, version, adjust_repos, **kw):
-    # note: when split packages for ceph land for SUSE,
-    # `kw['components']` will have those. Unused for now.
-    packages = ['ceph', 'ceph-radosgw']
+    packages = map_components(kw.get('components', []))
+
     pkg_managers.zypper_refresh(distro.conn)
-    pkg_managers.zypper(distro.conn, packages)
+    if len(packages):
+        pkg_managers.zypper(distro.conn, packages)
 
 
 def mirror_install(distro, repo_url, gpg_url, adjust_repos, **kw):
-    # note: when split packages for ceph land for SUSE,
-    # `kw['components']` will have those. Unused for now.
+    packages = map_components(kw.get('components', []))
     repo_url = repo_url.strip('/')  # Remove trailing slashes
     gpg_url_path = gpg_url.split('file://')[-1]  # Remove file if present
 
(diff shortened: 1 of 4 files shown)


ddiss:
FWIW, here's the test output...

hanson:~/ddiss/ceph-deploy/ceph_deploy/tests/unit/hosts/:[0]# nosetests -v test_suse.py
test_suse.TestSuseInit.test_choose_init_SLE_11 ... ok
test_suse.TestSuseInit.test_choose_init_SLE_12 ... ok
test_suse.TestSuseInit.test_choose_init_default ... ok
test_suse.TestSuseInit.test_choose_init_openSUSE_13_1 ... ok
test_suse.TestSuseMapComponents.test_invalid ... ok
test_suse.TestSuseMapComponents.test_valid ... ok

----------------------------------------------------------------------
Ran 6 tests in 0.002s

OK

trhoden:
OK to test

trhoden:
This looks really sane to me, and is a clear improvement for SUSE.

Sorry I did not jump in on the discussion on this in ceph-devel. I'd like to get a bit more consensus that it makes sense for ceph-deploy to point to distro packages by default for SUSE, but given that ceph.com doesn't build them anymore it is certainly a positive.

I'd rather we be consistent all across the board, with upstream vs. downstream by default, and the use of use_rhceph (if renamed to 'bundled' like you said) and things like that, but that may not be entirely possible.

trhoden:
please test this

trhoden:
...goes off to look up the magic plugin trigger phrase...

trhoden:
ok to test

trhoden:
apparently it's case sensitive.  yay.

trhoden:
@ddiss The style/linter checks failed here.  We will want them to pass before merging.

ddiss:
@trhoden thanks for the feedback!

For RHEL, the use_rhceph overrides the default (working) CentOS back-end, with a (working) RHEL back-end that installs the bundled packages.
For SUSE, the existing back-end is completely broken, so I'd prefer to fix it in-place rather than add a new SUSE-bundled back-end with this functionality.
Installation from external repositories using the SUSE back-end will continue to be possible via the install_mirror functionality.

Regarding the style issues, is there a way I can check that locally?

ddiss:
Superseded by #324 .

trhoden:
@ddiss as for the style issues -- yes, tox will do it.  If you just want to run the linter, you can run the flake8 command directly:

`flake8 --select=F --exclude=vendor ceph_deploy`

ddiss:
@trhoden I'll keep that in mind for next time - thanks for the follow up!

PR #313, opened 2015-06-30
Use argparse subparsers when subcommands have their own subcommands

trhoden:
This patchset changes the argparse config to use more subparsers to have much better help output and only list CLI options that are relevant to a given subcommand.  This may be best demonstrated by example.  Let's look at the `mon` subcommand.

`mon --help` before:

```
$ ceph-deploy mon --help
usage: ceph-deploy mon [-h] [--address [ADDRESS]] [--keyrings [KEYRINGS]]
                       {add,create,create-initial,destroy} [mon [mon ...]]

Subcommands:

create-initial
  Will deploy for monitors defined in `mon initial members`, wait until
  they form quorum and then gatherkeys, reporting the monitor status along
  the process. If monitors don't form quorum the command will eventually
  time out.

create
  Deploy monitors by specifying them like:

    ceph-deploy mon create node1 node2 node3

  If no hosts are passed it will default to use the `mon initial members`
  defined in the configuration.

add
  Add a monitor to an existing cluster:

    ceph-deploy mon add node1

  Or:

    ceph-deploy mon add node1 --address 192.168.1.10

  If the section for the monitor exists and defines a `mon addr` that
  will be used, otherwise it will fallback by resolving the hostname to an
  IP. If `--address` is used it will override all other options. Please
  note that only one node can be added at a time.

destroy
  Completely remove monitors on a remote host. Requires hostname(s) as
  arguments.

positional arguments:
  {add,create,create-initial,destroy}
  mon

optional arguments:
  -h, --help            show this help message and exit
  --address [ADDRESS]
  --keyrings [KEYRINGS]
                        concatenate multiple keyrings to be seeded on new
                        monitors
```
Besides being so long that users probably don't read it all, it's also misleading because it lists options at the top-level that do not apply to all subcommands, like `--keyrings` and `--address`.

`mon --help` after:
```
$ ceph-deploy mon --help
usage: ceph-deploy mon [-h] {add,create,create-initial,destroy} ...

Ceph MON Daemon management

positional arguments:
  {add,create,create-initial,destroy}
    add                 Add a monitor to an existing cluster:
                        	ceph-deploy mon add node1
                        Or:
                        	ceph-deploy mon add --address 192.168.1.10 node1
                        If the section for the monitor exists and defines a `mon addr` that
                        will be used, otherwise it will fallback by resolving the hostname to an
                        IP. If `--address` is used it will override all other options.
    create              Deploy monitors by specifying them like:
                        	ceph-deploy mon create node1 node2 node3
                        If no hosts are passed it will default to use the
                        `mon initial members` defined in the configuration.
    create-initial      Will deploy for monitors defined in `mon initial
                        members`, wait until they form quorum and then
                        gatherkeys, reporting the monitor status along the
                        process. If monitors don't form quorum the command
                        will eventually time out.
    destroy             Completely remove Ceph MON from remote host(s)

optional arguments:
  -h, --help            show this help message and exit
```

`mon add --help` before:
```
$ ceph-deploy mon add --help
usage: ceph-deploy mon [-h] [--address [ADDRESS]] [--keyrings [KEYRINGS]]
                       {add,create,create-initial,destroy} [mon [mon ...]]

Subcommands:

create-initial
  Will deploy for monitors defined in `mon initial members`, wait until
  they form quorum and then gatherkeys, reporting the monitor status along
  the process. If monitors don't form quorum the command will eventually
  time out.

create
  Deploy monitors by specifying them like:

    ceph-deploy mon create node1 node2 node3

  If no hosts are passed it will default to use the `mon initial members`
  defined in the configuration.

add
  Add a monitor to an existing cluster:

    ceph-deploy mon add node1

  Or:

    ceph-deploy mon add node1 --address 192.168.1.10

  If the section for the monitor exists and defines a `mon addr` that
  will be used, otherwise it will fallback by resolving the hostname to an
  IP. If `--address` is used it will override all other options. Please
  note that only one node can be added at a time.

destroy
  Completely remove monitors on a remote host. Requires hostname(s) as
  arguments.

positional arguments:
  {add,create,create-initial,destroy}
  mon

optional arguments:
  -h, --help            show this help message and exit
  --address [ADDRESS]
  --keyrings [KEYRINGS]
                        concatenate multiple keyrings to be seeded on new
                        monitors
```
I always found this difficult, as using `mon add --help` does not provide any new information about the `add` subcommand.  It just prints the same help as `mon --help`, including help for other subcommands and irrelevant options.

`mon add --help` after:
```
$ ceph-deploy mon add --help
usage: ceph-deploy mon add [-h] [--address [ADDRESS]] [mon [mon ...]]

positional arguments:
  mon

optional arguments:
  -h, --help           show this help message and exit
  --address [ADDRESS]
```

Now we are showing the user only the options that are relevant to the command they asked for help on.

The subcommands that got modified in this patchset are:
- calamari
- osd
- disk
- config
- mds
- mon
- rgw

I tried hard to make only one type of change here -- adding the subparsers.  There are other things that need to be fixed, but I purposely left them broken because they are covered under other open issues.

### DIFF
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -208,9 +208,6 @@ def prepare_disk(
     if zap:
         args.append('--zap-disk')
     if fs_type:
-        if fs_type not in ('btrfs', 'ext4', 'xfs'):
-            raise argparse.ArgumentTypeError(
-                "FS_TYPE must be one of 'btrfs', 'ext4' or 'xfs'")
         args.extend(['--fs-type', fs_type])
     if dmcrypt:
         args.append('--dmcrypt')
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -207,8 +207,6 @@ def prepare_disk(
         ]
     if zap:
         args.append('--zap-disk')
-    if fs_type:
-        args.extend(['--fs-type', fs_type])
     if dmcrypt:
         args.append('--dmcrypt')
         if dmcrypt_dir is not None:
@@ -217,6 +215,8 @@ def prepare_disk(
     args.extend([
         '--cluster',
         cluster,
+        '--fs-type',
+        fs_type,
         '--',
         disk,
     ])
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -767,45 +767,80 @@ def make_disk(parser):
     """
     Manage disks on a remote host.
     """
-    parser.add_argument(
-        'subcommand',
-        metavar='SUBCOMMAND',
-        choices=[
-            'list',
-            'prepare',
-            'activate',
-            'zap',
-            ],
-        help='list, prepare, activate, zap',
+    disk_parser = parser.add_subparsers(dest='subcommand')
+
+    disk_zap = disk_parser.add_parser(
+        'zap',
+        help='destroy existing partition table and content for DISK',
         )
-    parser.add_argument(
+    disk_zap.add_argument(
         'disk',
         nargs='+',
         metavar='HOST:DISK',
         type=colon_separated,
-        help='host and disk (or path)',
+        help='host and disk'
         )
-    parser.add_argument(
+
+    disk_list = disk_parser.add_parser(
+        'list',
+        help='List disk info from remote host(s)'
+        )
+    disk_list.add_argument(
+        'disk',
+        nargs='+',
+        metavar='HOST:DISK',
+        type=colon_separated,
+        help='remote host to list OSDs from'
+        )
+
+    disk_prepare = disk_parser.add_parser(
+        'prepare',
+        help='Prepare a disk for use as Ceph OSD by formatting/partitioning disk'
+        )
+    disk_prepare.add_argument(
         '--zap-disk',
         action='store_true', default=None,
         help='destroy existing partition table and content for DISK',
         )
-    parser.add_argument(
+    disk_prepare.add_argument(
         '--fs-type',
         metavar='FS_TYPE',
+        choices=['xfs',
+                 'ext4',
+                 'btrfs'
+                 ],
         default='xfs',
-        help='filesystem to use to format DISK (xfs, btrfs or ext4)'
+        help='filesystem to use to format DISK (xfs, btrfs, or ext4)',
         )
-    parser.add_argument(
+    disk_prepare.add_argument(
         '--dmcrypt',
         action='store_true', default=None,
         help='use dm-crypt on DISK',
         )
-    parser.add_argument(
+    disk_prepare.add_argument(
         '--dmcrypt-key-dir',
         metavar='KEYDIR',
         default='/etc/ceph/dmcrypt-keys',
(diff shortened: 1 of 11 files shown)


ceph_deploy/osd.py
```diff
@@ -706,46 +767,81 @@ def make_disk(parser):
...
+    disk_prepare.add_argument(
         '--fs-type',
         metavar='FS_TYPE',
+        choices=['xfs',
+                 'ext4',
+                 'btrfs'
+                 ],
         default='xfs',
-        help='filesystem to use to format DISK (xfs, btrfs or ext4)'
+        help='filesystem to use to format DISK',
```
alfredodeza: why are we removing the file system types supported? I know we have that as a common question coming up: "can I use $FS here?" or "I tried using $FS and it didn't work"

ceph_deploy/util/help_formatters.py
```diff
@@ -0,0 +1,11 @@
+import argparse
+
+
+class ToggleRawTextHelpFormatter(argparse.HelpFormatter):
+    """Inspired by the SmartFormatter at
+       https://bitbucket.org/ruamel/std.argparse
+   """
+    def _split_lines(self, text, width):
+        if text.startswith('R|'):
```
alfredodeza: this looks like a big gotcha that I didn't understand until I got the end of the review.  Would you mind adding a more thorough docstring with some examples so it is easier to parse/understand how it works?

ceph_deploy/osd.py
trhoden: I'm happy to put it back.  It's probably more useful that way.

I removed it because I added the `choices` field to enforce only the filesystems we allow. If you provide a bad one now, it will show:

```
$ ceph-deploy disk prepare --fs-type zfs
usage: ceph-deploy disk prepare [-h] [--zap-disk] [--fs-type FS_TYPE]
                                [--dmcrypt] [--dmcrypt-key-dir KEYDIR]
                                HOST:DISK [HOST:DISK ...]
ceph-deploy disk prepare: error: argument --fs-type: invalid choice: 'zfs' (choose from 'xfs', 'ext4', 'btrfs')
```

Previously the argument parser wouldn't kick this out, we would raise our own exception, which isn't necessary (we can fail sooner via argparse).

Howevever, you would still have to run the command with a bad fstype to see the list.  So I'll put it back.  And now that I've looked closer, I will remove the following redundant code:

```
if fs_type not in ('btrfs', 'ext4', 'xfs'):
            raise argparse.ArgumentTypeError...
```

ceph_deploy/util/help_formatters.py
trhoden: Yeah, I can do that.  I put a lot of the explanation in the original commit message too: https://github.com/trhoden/ceph-deploy/commit/4858f3a3b6f339004cfe55a104ce031884013fd6

trhoden:
@alfredodeza I think I addressed your two comments, and I made a couple of other changes to remove redundant code and unnecessary complexity around --fs-type handling.

PR #314, opened 2015-07-02
Set owner of directory and keyrings according to owner of /var/lib/ceph.

mbroz:
If Ceph is running under separate ceph user, it requires access
to various files under this user.
This patch tries to set owner of these files to the current owner
of base_dir (/var/lib/ceph).

If the ceph directory is owned by root, it expects old installation
and should work with old tools as well (avoids --setuser option).

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -123,7 +123,7 @@ def write_conf(cluster, conf, overwrite):
         raise RuntimeError(err_msg)
 
 
-def write_keyring(path, key):
+def write_keyring(path, key, uid=-1, gid=-1):
     """ create a keyring file """
     # Note that we *require* to avoid deletion of the temp file
     # otherwise we risk not being able to copy the contents from
@@ -133,27 +133,30 @@ def write_keyring(path, key):
     tmp_file.close()
     keyring_dir = os.path.dirname(path)
     if not path_exists(keyring_dir):
-        makedir(keyring_dir)
+        makedir(keyring_dir, uid, gid)
     shutil.move(tmp_file.name, path)
 
 
-def create_mon_path(path):
+def create_mon_path(path, uid=-1, gid=-1):
     """create the mon path if it does not exist"""
     if not os.path.exists(path):
         os.makedirs(path)
+        os.chown(path, uid, gid);
 
 
-def create_done_path(done_path):
+def create_done_path(done_path, uid=-1, gid=-1):
     """create a done file to avoid re-doing the mon deployment"""
     with file(done_path, 'w'):
         pass
+    os.chown(done_path, uid, gid);
 
 
-def create_init_path(init_path):
+def create_init_path(init_path, uid=-1, gid=-1):
     """create the init path if it does not exist"""
     if not os.path.exists(init_path):
         with file(init_path, 'w'):
             pass
+        os.chown(init_path, uid, gid);
 
 
 def append_to_file(file_path, contents):
@@ -161,6 +164,11 @@ def append_to_file(file_path, contents):
     with open(file_path, 'a') as f:
         f.write(contents)
 
+def path_getuid(path):
+    return os.stat(path).st_uid
+
+def path_getgid(path):
+    return os.stat(path).st_gid
 
 def readline(path):
     with open(path) as _file:
@@ -179,7 +187,7 @@ def listdir(path):
     return os.listdir(path)
 
 
-def makedir(path, ignored=None):
+def makedir(path, ignored=None, uid=-1, gid=-1):
     ignored = ignored or []
     try:
         os.makedirs(path)
@@ -189,24 +197,27 @@ def makedir(path, ignored=None):
         else:
             # re-raise the original exception
             raise
+    else:
+        os.chown(path, uid, gid);
 
 
 def unlink(_file):
     os.unlink(_file)
 
 
-def write_monitor_keyring(keyring, monitor_keyring):
+def write_monitor_keyring(keyring, monitor_keyring, uid=-1, gid=-1):
     """create the monitor keyring file"""
-    write_file(keyring, monitor_keyring)
+    write_file(keyring, monitor_keyring, 0600, None, uid, gid)
 
 
-def write_file(path, content, mode=0644, directory=None):
+def write_file(path, content, mode=0644, directory=None, uid=-1, gid=-1):
     if directory:
         if path.startswith("/"):
             path = path[1:]
         path = os.path.join(directory, path)
     with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, mode), 'w') as f:
         f.write(content)
+    os.chown(path, uid, gid)
 
 
 def touch_file(path):
@@ -289,7 +300,7 @@ def make_mon_removed_dir(path, file_name):
     shutil.move(path, os.path.join('/var/lib/ceph/mon-removed/', file_name))
 
 
-def safe_mkdir(path):
+def safe_mkdir(path, uid=-1, gid=-1):
     """ create path if it doesn't exist """
     try:
         os.mkdir(path)
@@ -298,9 +309,10 @@ def safe_mkdir(path):
             pass
         else:
             raise
+    else:
+        os.chown(path, uid, gid)
 
-
-def safe_makedirs(path):
+def safe_makedirs(path, uid=-1, gid=-1):
     """ create path recursively if it doesn't exist """
     try:
         os.makedirs(path)
@@ -309,6 +321,8 @@ def safe_makedirs(path):
             pass
         else:
             raise
+    else:
+        os.chown(path, uid, gid)
 
 
 def zeroing(dev):
--- a/ceph_deploy/hosts/common.py
+++ b/ceph_deploy/hosts/common.py
@@ -2,6 +2,7 @@ from ceph_deploy.util import paths
 from ceph_deploy import conf
 from ceph_deploy.lib import remoto
 from StringIO import StringIO
+from ceph_deploy.util import constants
 
 
 def ceph_version(conn):
@@ -15,6 +16,8 @@ def mon_create(distro, args, monitor_keyring, hostname):
     logger = distro.conn.logger
     logger.debug('remote hostname: %s' % hostname)
     path = paths.mon.path(args.cluster, hostname)
+    uid = distro.conn.remote_module.path_getuid(constants.base_path)
+    gid = distro.conn.remote_module.path_getgid(constants.base_path)
     done_path = paths.mon.done(args.cluster, hostname)
     init_path = paths.mon.init(args.cluster, hostname, distro.init)
 
@@ -30,7 +33,7 @@ def mon_create(distro, args, monitor_keyring, hostname):
     )
 
     # if the mon path does not exist, create it
-    distro.conn.remote_module.create_mon_path(path)
+    distro.conn.remote_module.create_mon_path(path, uid, gid)
 
     logger.debug('checking for done path: %s' % done_path)
     if not distro.conn.remote_module.path_exists(done_path):
@@ -44,8 +47,14 @@ def mon_create(distro, args, monitor_keyring, hostname):
         distro.conn.remote_module.write_monitor_keyring(
             keyring,
             monitor_keyring,
+            uid, gid,
         )
 
+        user_args = []
+        if uid != 0:
+            user_args = user_args + [ '--setuser', str(uid) ]
+        if gid != 0:
+            user_args = user_args + [ '--setgroup', str(gid) ]
         remoto.process.run(
             distro.conn,
             [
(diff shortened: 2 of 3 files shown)


osynge:
I really like your approach to "ceph_deploy/hosts/remotes.py" I prefer the safety aspects of the API, to my own chmod, chown,  but to reflect it better it should be nice to have some value validation, otherwise its just an obstruction to people getting the job done with my pull request.

Especially considering the more complex/limiting/learning curve api. I think some of the names could be improved but would not object to this being merged on this basis.

I do have to comment on your choice of complexities of uid's and gids in this patch though.

Doing it this way:

https://github.com/osynge/ceph-deploy/commit/c19e40b6407437f4dc1e0198d8d14401bb3e8ac2

Would allow for different uid/gid on different platforms assigned by external services, following local policy.

Hence I think this patch should be merged when it goes to "User name" / "Group name".

osynge:
or if it shows it can get the uid/gid from user and group names, rather than requiring a directory.

ceph_deploy/hosts/remotes.py
```diff
@@ -300,10 +311,11 @@ def safe_mkdir(path):
             raise
 
 
-def safe_makedirs(path):
+def safe_makedirs(path, uid=-1, gid=-1):
     """ create path recursively if it doesn't exist """
     try:
         os.makedirs(path)
+        os.chown(path, uid, gid)
```
osynge: this chown should be after the except.

ceph_deploy/hosts/remotes.py
```diff
@@ -289,10 +299,11 @@ def make_mon_removed_dir(path, file_name):
     shutil.move(path, os.path.join('/var/lib/ceph/mon-removed/', file_name))
 
 
-def safe_mkdir(path):
+def safe_mkdir(path, uid=-1, gid=-1):
     """ create path if it doesn't exist """
     try:
         os.mkdir(path)
+        os.chown(path, uid, gid)
```
osynge: this chown should be after the except.

ceph_deploy/hosts/remotes.py
```diff
@@ -195,18 +204,19 @@ def unlink(_file):
...
 
-def write_file(path, content, mode=0644, directory=None):
+def write_file(path, content, mode=0644, directory=None, uid=-1, gid=-1):
     if directory:
         if path.startswith("/"):
             path = path[1:]
         path = os.path.join(directory, path)
     with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, mode), 'w') as f:
         f.write(content)
+        os.chown(path, uid, gid)
```
osynge: indentation level

ceph_deploy/hosts/remotes.py
```diff
@@ -179,10 +187,11 @@ def listdir(path):
     return os.listdir(path)
 
 
-def makedir(path, ignored=None):
+def makedir(path, ignored=None, uid=-1, gid=-1):
     ignored = ignored or []
     try:
         os.makedirs(path)
+        os.chown(path, uid, gid);
```
osynge: after except

ceph_deploy/hosts/remotes.py
```diff
@@ -133,26 +133,29 @@ def write_keyring(path, key):
...
+        os.chown(done_path, uid, gid);
         pass
 
 
-def create_init_path(init_path):
+def create_init_path(init_path, uid=-1, gid=-1):
     """create the init path if it does not exist"""
     if not os.path.exists(init_path):
         with file(init_path, 'w'):
+            os.chown(init_path, uid, gid);
```
osynge: move this out of the "with file(..." context.

branto1:
lgtm
`Reviewed-by: Boris Ranto <branto@redhat.com>`

ceph_deploy/hosts/remotes.py
trhoden: Agreed.  It also needs to be moved for `create_done_path` immediately before this.

ceph_deploy/hosts/remotes.py
trhoden: I think I would put it in an `else:` clause after the except, so it is only run if there is not an exception. So this becomes:

```
try:
    os.makedirs(...)
except ....:
    ...
else:
    os.chown(...)
```

os.chown() is recursive, right? :)

ceph_deploy/hosts/remotes.py
trhoden: +1

ceph_deploy/hosts/remotes.py
trhoden: same thing about the `else:` bit here and the next one.

trhoden:
>I really like your approach to "ceph_deploy/hosts/remotes.py" I prefer the safety aspects of the API, to my own chmod, chown, but to reflect it better it should be nice to have some value validation, otherwise its just an obstruction to people getting the job done with my pull request.

@osynge can you clarify what you mean here, as far as value validation?  Are you talking about the uid/gid values?  I don't follow the bit about being an obstruction.

>Doing it this way:
>
>osynge@c19e40b
>
>Would allow for different uid/gid on different platforms assigned by external services, following local policy.
>
>Hence I think this patch should be merged when it goes to "User name" / "Group name".

So you are against reading the user and group names from the /var/lib/ceph directory?  It seems like it does handle different uid/gid values per host by reading whatever it happens to be on that directory.

> or if it shows it can get the uid/gid from user and group names, rather than requiring a directory.

Where do you think it should get the user/group names from in this case?

osynge:
I will answer both parts in different resonances.

osynge:
"value validation" and preferences.

Why I prefer this approach to having a separate chmod and chown function, is we cant restrict which files are "chown" and "chmod" as thier is no idea in the code what they are doing.

I was hoping we can make some rules in the remote library that could provide some validation they are not say changing the permission of say /boot/kernel.

I am not 100% convinced this is practical but it would be nice if we could do some thing in this direction. But I think this is probably out of scope for this patch, but if the author has some ideas I should love to here them.

mbroz:
The idea to use owner of /var/lib/ceph is by Sage, discussed here http://article.gmane.org/gmane.comp.file-systems.ceph.devel/25671

The rest is just nitpicking, I really do not care where chown is called :)
IMHO chown should fail the same way the create fails (ignoring it means problem later which is much more hard to spot). If you want it changed, I'll submit new patch once I am back online.

Also I would not expect there will be other user than fixed "ceph" user here because there is a lot of not easily configurable dependences (udev rules need to assign OSD block devices, see our approach here https://github.com/ceph/ceph/commit/5185e96b0289b02dc29a645e3fa0dc75e1043f88 etc).

osynge:
"uid and gid Vs username and groupname"

Generally in my experience data centres are happy to define a range of UID/GID for a project, but will never agree what the values are. Conversely data centres are happy to give out "user names" and "group names" that are the same at many sites.

SUSE and I am sure some other distributions cannot guarantee a UID/GID is free before installation of a cluster at a site, and due to this they fall into line with my experience. Redhat I belive has been more restricitive in the past with thier UID/GID ranges and can give ceph a single UID/GID for all sites, but I doubt this can be achieved for all distributions.

As a ceph-deploy contributor I am interested in using abstractions to solve cross site and cross distribution issues in the simplest way. Using User names and group names is vastly easier when we put the customer in mind.

I think directory paths are not a good way to define user, as they are not a good entry point to defining this as some thing has to create them, making ceph-deploys behaviour undefined. Better would be make a method for each "distro" to get the username and group name for each service directory, or thing and we are done and never have to face nasty UID/GID complexities in cross site cross distribution problems.

I see 2 ways to make my concerns about using UID/GID go away:
 * user name / group name mapping to uid/gid; happening at the client in each method.
 * user name / group name mapping to uid/gid; happening at the client and passed back to the server, which then calls the appropriate commands with the appropriate UIG/GID.

I prefer the first approach (as its more self contained in a library, easier to use, and prevents any developer doing the wrong thing with UID/GID) but either is acceptable.

osynge:
Thanks for your reasoning to using directories as the source. I think sage is wrong here, that’s not to say he is anything but usually right.

osynge:
In reply to:

(udev rules need to assign OSD block devices, see our approach here ceph/ceph@5185e96 etc).

Please note this is "user name" and "group name" level abstraction also. 

Yes and I see this as the sort of thing that in my opinion should be templates from autotools, as "user names" and "group names" are site / distro choices, not upstreams, but they can put in sensible defaults.

I want to avoid any nasty stuff with UID's and GID's ideally as it gets nasty very very easily.

osynge:
@liewegas should we discuss my objection here to the use of directories to gather UID GID information here or on the ceph-devel mailing list?

osynge:
In SUSE the "wwwrun" username and "www" is what we will run the fastcgi rgw application under.

As for civetweb I don’t yet know, maybe the same user, but their will be more issues with that in packaging. I am hoping their will be no issues getting the rest of ceph to run as user and group "ceph"

mbroz:
Just FYI, I actually run ceph MON/OSD under ceph user, installed by ceph-deploy with the patch in this PR.

osynge:
@mbroz that is indeed interesting. 

@smithfarm I guess we are to late to get this into the release, but its tempting, but please note my comments to mbroz about how I at least dont want to see UID/GID in our code base.

l-mb:
Well, picking up the uid/gid from /var/lib/ceph on the target node (after the ceph package has been installed) should make sense, right? I'd imagine SUSE creates /var/lib/ceph in %postin, adds/reuses a ceph user/group there.

osynge:
* I am not bothered about it working (sure it does).
 * I am bothered about it not working, and more scary where it may evolve, and how these numbers can be handled.
 * UID's and GID's are fine as long as no one ever assumes they are cross site and cross platform.
 * user names and group names from my experience can reasonably be considered cross site and cross platform.
 * I don't like "magic" in code and expect a CLI to have command line override from all "magic", and auto detection, and at least log where the "magic" comes from.

@l-mb since you disagree, I drop my argument that this is bad practice.

ceph_deploy/hosts/remotes.py
osynge: Personally I would have it independent of the except but my wording was unclear, but the else is fine too. Is os.chown() recursive? I did not think so. and looking at the docs it makes no specific mention of this. https://docs.python.org/2/library/os.html

mbroz: os.chown() is not recursive. But it is not easy to implement it recursive (I tried and it has some problems with remote specification, locally it is just a few lines). But is is enough that the last dir has proper access.
(Moreover, I did not find any use that it actually creates more than actual dir. Maybe the whole makedirs is overkill there.)

l-mb:
@osynge That's why I wrote *on the target node*; that doesn't imply they are the same across nodes (much less sites), which we clearly cannot guarantee.

osynge:
@l-mb I understood your perspective that it will work when done locally before. I have no doubt this will work.

My previous view was based on practices developed for mass deployment with Grid Computing after the community I worked with suffered before particularly at sites with very large numbers of UID/GID's in common use (at a global scale of say all experimental particle physicists globally), often integrating with very large numbers of "local users", (Ie all UK or German physicists at doctorate or above at the the national level) where we even had problems at one point even with limitations of available max numbers for UID/GID's (which may have been resolved in kernels of the past 10 years) and how to handle multiple identity providers site YP/NIS, and kerberos etc. 

These extreme cases may not be relevant to ceph as they where to batch queues and clusters where UID's and GID's may not HAVE to be synchronised as was policy in such sites as I used to work with. Since you are more an expert than I, in the open source world, and I trust your judgement, if UID/GID Vs username/groupname don't concern you, as having nasty side effects later in development, I am just worrying too much based on having working in funny environments with extreme situations.

mbroz:
I fixed the indentation and moved chown to else after except label. Please have a look.

alfredodeza:
this looks good to me

osynge:
@l-mb  please review this code also.
@mbroz I am happy with everything but the uid/gid thing.

Since I am the only one who is concerned that ceph-deploy is not defining the user/group, and the only one who thinks working with UID/GID is storing up trouble and the only one who objects to it not being over-ridable, and the only one who thinks it needs logging and error handling. I am not going to object to this being merged following the latest update.

trhoden:
ok to test

trhoden:
This looks good to me. The directory ownership is taken care of by the packages, and ceph-deploy is intended to work with the packages.

The name -> id mappings happen individually on each node, and there are no hardcoded names or IDs that I see.

I will merge this in once we get passing tests.  Looks like Jenkins is having issues at the moment.

trhoden:
Got a better test run here: http://jenkins.ceph.com/job/ceph-deploy-pull-requests/1698/

trhoden:
I'm not sure why the tests are triggering this:

```
[ceph_deploy][ERROR ]   File "/home/jenkins-build/build/workspace/ceph-deploy-pull-requests/ceph-deploy/ceph_deploy/util/decorators.py", line 69, in newfunc
[ceph_deploy][ERROR ]   File "/home/jenkins-build/build/workspace/ceph-deploy-pull-requests/ceph-deploy/ceph_deploy/cli.py", line 174, in _main
[ceph_deploy][ERROR ]   File "/home/jenkins-build/build/workspace/ceph-deploy-pull-requests/ceph-deploy/ceph_deploy/admin.py", line 40, in admin
[ceph_deploy][ERROR ]     0600,
[ceph_deploy][ERROR ] TypeError: write_file() takes at least 5 arguments (3 given)
[ceph_deploy][ERROR ] 
```

`write_file()` has optional arguments, it should be fine to not pass them (though they have to passed by position, not name. This is an `execnet` limitation.  The code looks fine to me -- I don't know why it would say that 5 args are required.  It should be 2 required, with 4 optional.

When I get to it, I'll run this code manually and see if it works.  This was triggered by the `ceph-deploy admin` command.

trhoden:
@mbroz I have one set of changes to ask for, to get the failed test to past.

The test at `tests.test_cli_admin.test_write_keyring` overwrites the default values for the `write_file` method's optional parameters.  These needs to be updated to cover the new params as follows:

```
$ git diff
diff --git a/ceph_deploy/tests/test_cli_admin.py b/ceph_deploy/tests/test_cli_admin.py
index aeb0778..57bfe31 100644
--- a/ceph_deploy/tests/test_cli_admin.py
+++ b/ceph_deploy/tests/test_cli_admin.py
@@ -44,7 +44,7 @@ def test_write_keyring(tmpdir):

     distro = MagicMock()
     distro.conn = MagicMock()
-    remotes.write_file.func_defaults = (str(tmpdir),)
+    remotes.write_file.func_defaults = (0644, str(tmpdir), -1, -1)
     distro.conn.remote_module = remotes
     distro.conn.remote_module.write_conf = Mock()
```

If you can make that update, we should be good to go.

mbroz:
@trhoden Patch is folded in.

PR #317, opened 2015-07-06
services abstraction to simplify code.

osynge:
While developing a rgw abstraction for fastcgi radosgw I found starting
and stopping services code was set with conditionals duplicated 
This code hopes to solve this issue in one library function.

### DIFF
--- a/ceph_deploy/rgw.py
+++ b/ceph_deploy/rgw.py
@@ -9,7 +9,7 @@ from ceph_deploy import hosts
 from ceph_deploy.util import system
 from ceph_deploy.lib import remoto
 from ceph_deploy.cliutil import priority
-
+from ceph_deploy.util import services
 
 LOG = logging.getLogger(__name__)
 
@@ -56,6 +56,7 @@ def create_rgw(distro, name, cluster, init):
             os.path.join(keypath),
         ]
     )
+    entity_name = 'client.bootstrap-rgw'
     if returncode > 0 and returncode != errno.EACCES:
         for line in stderr:
             conn.logger.error(line)
@@ -70,7 +71,7 @@ def create_rgw(distro, name, cluster, init):
             [
                 'ceph',
                 '--cluster', cluster,
-                '--name', 'client.bootstrap-rgw',
+                '--name', entity_name,
                 '--keyring', bootstrap_keyring,
                 'auth', 'get-or-create', 'client.{name}'.format(name=name),
                 'osd', 'allow *',
@@ -83,31 +84,13 @@ def create_rgw(distro, name, cluster, init):
     conn.remote_module.touch_file(os.path.join(path, 'done'))
     conn.remote_module.touch_file(os.path.join(path, init))
 
-    if init == 'upstart':
-        remoto.process.run(
-            conn,
-            [
-                'initctl',
-                'emit',
-                'radosgw',
-                'cluster={cluster}'.format(cluster=cluster),
-                'id={name}'.format(name=name),
-            ],
-            timeout=7
-        )
-    elif init == 'sysvinit':
-        remoto.process.run(
-            conn,
-            [
-                'service',
-                'ceph-radosgw',
-                'start',
-            ],
-            timeout=7
-        )
-
-    if distro.is_el:
-        system.enable_service(distro.conn, service="ceph-radosgw")
+    init_obj = init_system(connection = distro.conn,
+            init_type = distro.choose_init(),
+            service_name_mapping = distro.service_mapping)
+    try:
+        init_obj.start("rgw",[entity_name])
+    except init_exception_service:
+        LOG.error("Failed starting rgw:%s" % (entity_name))
 
 
 def rgw_create(args):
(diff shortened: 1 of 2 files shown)


osynge:
#318 should be considered a dependency for this code, did not realise they where not independent until rereading.

trhoden:
Thanks for taking a look at this.

We've long wanted to redo how we were handling services, and knew the right time to do it would be when we were adding support for systemd.  There are some things in here I like.

The classes `init_system_systemd` and `init_system_sysV` are on the right track. We would need to add one for upstart, too.  I find the abstraction of these classes done by `init_system` to be unnecessary, though. I think where we would like to go with this is to have the assignment of the init_system "handler" be done when the connection to the remote node is made.  e.g. here: https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/hosts/__init__.py#L71.

There would be logic in each module's `choose_init` call that either returns an instance of `init_system_{system}` or a name saying the init type, then we create an instance of the class.  In this way, calls for all hosts can perhaps be done with: `distro.init.enable()` and `distro.init.start()` instead of the current "if systemd: do this, else: do that".

However, perhaps this is too simplistic for systemd, as there is the complication of creating service names like `ceph@osd` or however those are constructed (and I see you did some work to handle).  I think it may still be able to be done that way, however.

The other way we have tossed around doing this is instead of having the rgw create details be in `ceph_deploy/rgw.py`, we make it look more like `mon` is done, with each host module having a `rgw` directory (like `hosts/centos/rgw`) so that we can put more distro specific info in there, rather than having lots of if/else statements.  I think this last part I am mentioning can be done regardless, and also negates the need for #318 .

osynge:
Learning Upstart is not going to happen for me, so is this patch dead now?

Even though when restarting 3 services that means 3X20 lines of decode almost duplication, along with all the duplication in osd and mon methods for their service management?

In favour you propose, all functions be made to duplicate for debian, ubuntu, centos, fedora, and suse ?

I am confused?

Cant you write the bug fix for upstart and work with me ?

osynge:
systemd is not distro specific, look in the code base and it shows how this is a dimension to operating systems dependent on version, I fear your argument in favour of an unimplemented solution would lead to massive code duplication.

osynge:
You should know that using modules + using functions and function level variables forbids multi threading or concurrency.

Lots of things can be done to hide that this is a bad idea, such as to instantiate the modules you use as objects, but they dont get away with being much simpler than an object with a fa card.

osynge:
You can get around the above failing of using a Facade_pattern to hide modules rather than objects, by importing modules as objects. Then the only disadvantage is the complexity of the import path, and the enforcement of relative path names, compared the simplicity of doing it here in one file.

osynge:
Please note the example usage patch is not intended to be merged.

trhoden:
Okay, there's a bunch of stuff here...

> Learning Upstart is not going to happen for me, so is this patch dead now?

Absolutely not dead

>Even though when restarting 3 services that means 3X20 lines of decode almost duplication, along with all the duplication in osd and mon methods for their service management?
>
>In favour you propose, all functions be made to duplicate for debian, ubuntu, centos, fedora, and suse ?
>
>I am confused?

I probably said something poorly.  I don't want lots of code duplication either.  Looking at your sample usage patch helped quite a bit.  I still think things could be made a lot simpler by just instantiating an instance of `init_system_{system}` in https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/hosts/__init__.py, so that is available later as `distro.init`.  Then your example usage patch would use things like `distro.init.start()` and `distro.init.stop()`.  I think we are ending up in the same place, just getting there slightly differently.

>Cant you write the bug fix for upstart and work with me ?

Of course.  I never intended to imply that this was dead without upstart.  I was merely noting that upstart would need to be implemented as well.

>systemd is not distro specific, look in the code base and it shows how this is a dimension to operating systems dependent on version, I fear your argument in favour of an unimplemented solution would lead to massive code duplication.

No it is not distro specific.  Which is why having the classes for dealing with systemd vs sysv makes sense.  I would do the same thing.  What *is* distro specific is defining the mapping of which distro versions use which init system.  such that centos 6 uses sysv, centos7 uses systemd.  Ubuntu 12.04 and 14.04 uses upstart, but 16.04 will use systemd, etc.

>Using Facade_pattern to hide modules rather than objects, is not allowing more than one concurrent implementation which is a very big architectural fail. Penalties include, no concurrence, no model comparing, in summary I think the current way of implementing Facade_pattern is bad practice that ceph-deploy probably wont ever migrate out of.
>
>You should know that using modules + using functions and function level variables forbids multi threading or concurrency. Hence even just having two open connections to both the mon and the rgw at the same time when they are on different operating systems might prove buggy.

Frankly I am not prepared to debate this topic.  You may be right, and I am certainly not saying that the current situation could not be improved.  That's the nature of software, right?  I think the important thing for us to note right now is that moving away from a facade paradigm, if desired, would be a longer term project that would be worked in a long-running branch (not master) and would be as part of a greater "new and improved" ceph-deploy.  I.e. a 2.0 version or something.  between the the facade argument and the desire to move to MVC, these both sound like major restructuring efforts that would not go into the current 1.5.x stable series.  That does not mean they are without merit.

osynge:
I just added a bug fix to the systemd service status function.

osynge:
Just added a restart function to the façade.

PR #318, opened 2015-07-06
Service mapping overrides and function.

osynge:
Each operating system calls services differently. This function gives
each operating system its own overrides for the "ceph" names for services.

### DIFF
--- a/ceph_deploy/hosts/centos/__init__.py
+++ b/ceph_deploy/hosts/centos/__init__.py
@@ -28,3 +28,12 @@ def choose_init(module):
 
 def get_packager(module):
     return pkg_managers.Yum(module)
+
+
+def service_mapping(service):
+    """
+    Select the service name
+    """
+    service_mapping = { "apache" : "httpd",
+        "ceph-rgw" : "ceph-rgw" }
+    return service_mapping.get(service,service)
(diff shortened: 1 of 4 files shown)


trhoden:
I'm not sure I see the need for this.  ceph-deploy does not support apache, so that part is moot.  There is indeed a difference between `radosgw` and `ceph-radosgw` on RPM and non-RPM systems, but those details are (hopefully) handled already.  Though now that I think about it, I know Debian is broken (http://tracker.ceph.com/issues/12170) because it equates `radosgw` the service with upstart, which doesn't work on Debian.

So, on second thought, we do need something along these lines to fix `12170`.  I think putting in this mapping makes sense, but here is what we need to see:

* drop all the apache references
* the key in the mappings should be 'rgw'
* `ceph-rgw` doesn't exist.  -EL distros are `ceph-radosgw`, and DEB distros are `radosgw`.
* make sure the mapping is in place for the rhel module as well
* Add another commit that replaces the hardcoded service names at https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/rgw.py#L92 and https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/rgw.py#L103 with the string from a call to distro.service_mapping("rgw")
* Add tests making sure that all distro return the correct service name for rgw.

How does that sound?

osynge:
Sure but, 1 I would like to keep the references for suse as these are used but changed to "rgw".

I cant test rhel, and centos, due to tools limitations, only fedora 22.

osynge:
I wont make the changes until I have the patch that depends also building with the new value though.

osynge:
@trhoden  rebased to master to make this clearer for https://github.com/ceph/ceph-deploy/pull/355

PR #319, opened 2015-07-06
Add remotes.chmod and wip_remotes.chown

osynge:
When running components of ceph as non root, these
functions become important.

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -5,7 +5,8 @@ import os
 import shutil
 import tempfile
 import platform
-
+import pwd
+import grp
 
 def platform_information(_linux_distribution=None):
     """ detect platform information from remote host """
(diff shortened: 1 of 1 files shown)


osynge:
@alfredodeza  could you please review this.

trhoden:
This looks good to me.  Please amend the commit title to say `remotes.chown` instead of `wip_remotes.chown`.  With that done:

`Reviewed-by: Travis Rhoden <trhoden@redhat.com>`

I'll wait a bit to see if @alfredodeza has anything to add.

ktdreyer:
We're adding function definitions with no uses or tests?

osynge:
Good catch thanks trhoden.

osynge:
@ktdreyer we are gathering some dependency for some big patches and reviewing them. we have not talked about merging, and I would imagine a collection of these pull requests will all be merged into change. Please file a bug on lack of tests if this ever gets merged, this time I suggest it will save not just me but also  trhoden lots of work if we change the order just slightly.

trhoden:
@ktdreyer I was also thinking it might be useful for #314 but now that I look at it, it might not be.

There's not a whole lot to test with these...

@osynge Is there any reason I should hold off on merging this? it looks ready to go to me once the commit title is changed.

Please do me a favor and also take a look at what @mbroz has done in #314, since I think you are both looking into the same thing.

osynge:
No reason to hold off merging it that I know of.

trhoden:
@osynge Is this PR still something you are looking to get in?  Or did the changes in #314 change the rationale behind this?

osynge:
At this moment I am not desperate to get this in.

I think this patch will now be modified to extend https://github.com/ceph/ceph-deploy/pull/314 

But I don’t think it will get to it before a weeks holiday.

liewegas:
lgtm

alfredodeza:
I don't see anything that is calling these functions. Feel free to re-open when you rebase and rework them

PR #320, opened 2015-07-07
DNM : Model View Controlers make code far simpler

osynge:
This is the start of an MVC pattern using an sqlite in memmory database and accessing it in pythoin with sqlalchamy.

### DIFF
--- a/ceph_deploy/mdl_loader.py
+++ b/ceph_deploy/mdl_loader.py
@@ -1,0 +1,75 @@
+import rdbms as model
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+import logging
+
+from v_pools import mdl_updater
+
+# use in-memory database
+databaseConnectionString = 'sqlite://'
+
+
+class mdl_controler(object):
+    def __init__(self,args,**kwargs):
+        connection = kwargs.get('connection', 'sqlite://')
+        sql_log = kwargs.get('sql_log', False)
+        self.args = args
+        self.log = logging.getLogger("db_controler")
+
+        self.engine = create_engine(connection, echo=sql_log)
+        model.init(self.engine)
+        self.SessionFactory = sessionmaker(bind=self.engine)
+
+
+    def cfg_update(self,cfg):
+        Session = self.SessionFactory()
+        fsid = cfg.safe_get('global', 'fsid')
+        matching_clusters = Session.query(model.Cluster).\
+            filter(model.Cluster.identifier == fsid).\
+            filter(model.Cluster.name == self.args.cluster)
+        if matching_clusters.count() == 0:
+            cluster = model.Cluster(
+                identifier = fsid,
+                name=self.args.cluster
+                )
+            Session.add(cluster)
+            Session.commit()
+            matching_clusters = Session.query(model.Cluster).\
+                filter(model.Cluster.identifier == fsid)
+        cluster = matching_clusters.one()
+
+        initial_mon = cfg.safe_get('global', 'mon_initial_members')
+        for host in initial_mon.split(','):
+            matching_node = Session.query(model.Node).\
+                filter(model.NodeHostName.hostname == host).\
+                filter(model.Node.id == model.NodeHostName.node)
+            if matching_node.count() == 0:
+                node = model.Node()
+                Session.add(node)
+                Session.commit()
+                node_hostname = model.NodeHostName(
+                    node=node.id,
+                    hostname=host
+                    )
+                Session.add(node_hostname)
+                Session.commit()
+                matching_node = Session.query(model.Node).\
+                    filter(model.NodeHostName.hostname == host).\
+                    filter(model.Node.id == model.NodeHostName.node)
+            # now we should make sure all nodes have a service instance associated.
+            for wnode in matching_node:
+                matching_mon = Session.query(model.InstanceMon).\
+                    filter(model.InstanceMon.node == wnode.id)
+                if 0 == matching_mon.count():
+                    mon_instance = model.InstanceMon(
+                        node = wnode.id,
+                        cluster = cluster.id,
+                        )
+                    Session.add(mon_instance)
+                    Session.commit()
+
+
+    def mon_update(self):
+        view = mdl_updater(SessionFactory=self.SessionFactory,
+            args= self.args)
+        view.mon_update()
--- a/ceph_deploy/v_pools.py
+++ b/ceph_deploy/v_pools.py
@@ -1,0 +1,243 @@
+import rdbms as model
+import logging
+import json
+from ceph_deploy import hosts
+from ceph_deploy.lib import remoto
+
+# pool tools
+LOG = logging.getLogger(__name__)
+
+def pool_list(conn):
+    stdout, stderr, rc = remoto.process.check(
+        conn,
+        [
+            'ceph',
+            '-f',
+            'json',
+            'osd',
+            'lspools'
+            ],
+        )
+    if rc != 0:
+        LOG.debug("stdout=%s" % (stdout))
+        LOG.debug("stderr=%s" % (stderr))
+        LOG.debug("rc=%s" % (rc))
+        return None
+    auth = json.loads("".join(stdout).strip())
+    return auth
+
+def pool_add(conn, name, number):
+    stdout, stderr, rc = remoto.process.check(
+        conn,
+        [
+            'ceph',
+            'osd',
+            'pool',
+            'create',
+            name,
+            str(number)
+            ],
+        )
+    if rc != 0:
+        LOG.debug("stdout=%s" % (stdout))
+        LOG.debug("stderr=%s" % (stderr))
+        LOG.debug("rc=%s" % (rc))
+        return None
+
+def pool_del(conn, name):
+    stdout, stderr, rc = remoto.process.check(
+        conn,
+        [
+            'ceph',
+            'osd',
+            'pool',
+            'delete',
+            name,
+            name,
+            '--yes-i-really-really-mean-it'
+            ],
+        )
+    if rc != 0:
+        LOG.debug("stdout=%s" % (stdout))
+        LOG.debug("stderr=%s" % (stderr))
+        LOG.debug("rc=%s" % (rc))
+        return None
+
+
+class mdl_updater(object):
+    def __init__(self,**kwargs):
+        self.SessionFactory = kwargs.get('SessionFactory', None)
+        self.args = kwargs.get('args', None)
+        self.distro = None
+        self.clusterId = None
+
+    def connect(self,clusterId):
+        self.clusterId = clusterId
+        LOG.info("clusterId=%s" % (clusterId))
+        Session = self.SessionFactory()
+        cluster_qry = Session.query(model.Cluster).\
+            filter(model.Cluster.identifier == clusterId)
+        if cluster_qry.count() == 0:
+            LOG.warning("Cluster '%s' does not have any known mon services" % (
+                clusterId,
+                ))
+            return False
+        cluster = cluster_qry.one()
+        possible_mons = Session.query(model.NodeHostName).\
+            filter(model.InstanceMon.cluster == cluster.id).\
+            filter(model.InstanceMon.node == model.Node.id).\
+            filter(model.NodeHostName.node == model.Node.id)
+        if possible_mons.count() == 0:
+            LOG.warning("Cluster '%s' with fsid '%s' does not have any known mon services" % (
+                cluster.name,
+                cluster.identifier
+                ))
+            return False
+        distro = None
+        # find the first mon we can connect too,
+        # it does not matter which
+        for mon in possible_mons:
+            try:
+                distro = hosts.get(mon.hostname, username=self.args.username)
+            except:
+                # try next host
+                LOG.warning("Failed to connect to '%s' with username '%s'\n" % (
+                    mon.hostname,
+                    self.args.username,
+                    ))
+                return False
+            break
+        distro = hosts.get(mon.hostname, username=self.args.username)
+        if distro == None:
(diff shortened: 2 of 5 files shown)


ceph_deploy/v_pools.py
```diff
@@ -0,0 +1,243 @@
...
+            ],
+        )
+    if rc != 0:
+        LOG.debug("stdout=%s" % (stdout))
+        LOG.debug("stderr=%s" % (stderr))
+        LOG.debug("rc=%s" % (rc))
+        return None
+
+
+class mdl_updater(object):
```
osynge: This class and its methods are not very MVC.

Better separation between the Model View's and controller should be done here to be a good example.

I see this code as being split into 2 views one for getting the sate, and one for updating the state of the cluster.

The controller would be elsewhere.

ceph_deploy/mdl_loader.py
```diff
@@ -0,0 +1,75 @@
...
+        sql_log = kwargs.get('sql_log', False)
+        self.args = args
+        self.log = logging.getLogger("db_controler")
+
+        self.engine = create_engine(connection, echo=sql_log)
+        model.init(self.engine)
+        self.SessionFactory = sessionmaker(bind=self.engine)
+
+
+    def cfg_update(self,cfg):
```
osynge: This method should be moved into a view, as its more view than controller.

liewegas:
I'm not sure MVC is appropriate for ceph-deploy.  One of its nice properties is its simplicity: it has no cluster state except for $cluster.* in the current directory, and those are all plaintext flat files that you can copy to/from remote hosts (manually for via simple commands).

osynge:
@liewegas  I disagree I see state in:

The local config file.
The nodes config files.
The nodes installed packages.
The command line used.
The cluster, 
The services running.
The operating systems running.
The pools that exist.
If the server is up.

Having this in one data structure, (sqlalchamy or simple python dicts) would allow us to make the code that establishes these state gathering functions to exist once rather than be duplicated.

We can then derive what to do:

If the operation should succeed.
One way to query all these different components state.
One way to compare desired and current state.

We can also use the state gathering code, to validate the state transition, and also inject state to make unit tests easily.

Personally I think MVC adoption makes the code a lot simper, in terms of branching and duplication, as views can be made to focus on only one thing, where as most ceph-deploy functions are mix many features and include conditionals, of if in this state or on this OS, where as they could be split into functions that are much smaller, that either do some thing (Views) or decide what to do (controller).

MVC patterns can be nested, and can share models. I dont care about the implementation.

Just my opinion :)

osynge:
As a rather clear example of how a MVC would benefit these 7 functions:

https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/osd.py#L344
https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/osd.py#L269
https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/rgw.py#L140
https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mon.py#L232
https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mon.py#L370
https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mon.py#L394
https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/install.py#L81

Which all iterate over hosts, (and it happens all over ceph-deploy) rather than establishing which hosts they should act upon, when for example they have possibly been processed or partially processed on a previous execution.

liewegas:
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/osd.py#L344
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/osd.py#L269
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/rgw.py#L140
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mon.py#L232
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mon.py#L370
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mon.py#L394
> https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/install.py#L81

The whole point is that ceph-deploy does what the admin tells it to do, 
and nothing else.  Of course it iterates over teh list of hosts or devices 
the admin tells it to install on.  It's not (and isn't meant to be) 
something more omnicient.

The main thing I can think of in ceph-deploy that is state-dependent is 
where it detects the distro and acts accordingly.  We'll have to add a bit 
more (at least temporarily) to deal with the ceph version so that we do 
sysvinit for hammer and systemd for >= infernalis.  That is the kind of 
state that we *shouldn't* persist, though, because it can change at 
any time for any number of reasons, and ceph-deploy should not make 
assumptions.

osynge:
Why is it not be possible to take $cluster.* in the current directory, and just run a single command that sets up the monitors and anything else it can gather from the $cluster.conf,  file and command line and apply it to the nodes it will mange and find in the cluster.

Ie why does any one ever have to know the difference between:

    ceph-deploy mon create

and 

    ceph-deploy mon create-initial

A clear failing of ceph-deploy atm caused in the most part by ceph-deploy not being MVC is that you need to run complex sequences of ceph-deploy command to setup a cluster, in defined orders, when actually all you wanted to do was run the change from the last time you ran a test cluster:

   ceph-deploy rgw create $HOSTNAME

as you already had setup ceph.conf from a previous test cluster install.

liewegas:
On Wed, 5 Aug 2015, Owen Synge wrote:
> Why is it not be possible to take $cluster.* in the current directory, and just run a single command that sets up the monitors and anything else
> it can gather from the $cluster.conf, file and command line and apply it to the nodes it will mange and find in the cluster.
> 
> Ie why does any one ever have to know the difference between:
> 
> ceph-deploy mon create
> 
> and
> 
> ceph-deploy mon create-initial

IMO if that's what you want then use puppet or chef or whatever--that's 
what those tools are designed to do.

> A clear failing of ceph-deploy atm caused in the most part by ceph-deploy not being MVC is that you need to run complex sequences of ceph-deploy
> command to setup a cluster, in defined orders, when actually all you wanted to do was run the change from the last time you ran a test cluster:
> 
> ceph-deploy rgw create $HOSTNAME
> 
> as you already had setup ceph.conf from a previous test cluster install.

Not sure what you mean... the above should be all you need to do.  Unless 
you've gone and modified ceph.conf everywhere, in which case you can 
refresh first (or not--it usually won't matter).

osynge:
Ok I try to be clearer:

Assuming you have a freshly wiped test cluster, with authorisation and repositories setup, but have previously setup a ceph.conf file, why do you once again have to run these commands in the correct order:

    ceph-deploy install ceph-node2 ceph-node3 ceph-node4
    ceph-deploy mon create-initial
    ceph-deploy --overwrite-conf rgw create ceph-node3:onr

When all the data needed to define the commands is already available in ceph.conf.

If you had a model, that gathered the desired state, a way to establish the current state, a way to compare states, and a way to apply states, this could just be:

    ceph-deploy apply-cfg

Essentially this is the strength of MVC. Also if you ran a redundant reapply ceph-deploy could simply gather the state and be able to tell you no operations are needed.

This is not hard to do, with the MVC design pattern.

osynge:
Another clear example of how ceph-deploy an make a huge leap in user friendliness by moving to MVC:

All commands can have a flag:

    --overwrite-conf

Without this flag many operations should fail, if the config remotely does not match the local config, and in a user friendly way should fail before any other operations have been attempted.

With MVC, you have a model, that gathered the desired state, a way to establish the current state, a way to compare states. So before any operation is started you can check if you do need the flag.

Without MVC, you are likely to only be able to unify the code to check the need for the flag to allow configuration over writing at some point when you attempt to overwrite the config, which may or may not be the most user friendly / or code friendly place in the config setup.

liewegas:
On Wed, 5 Aug 2015, Owen Synge wrote:
> Ok I try to be clearer:
> 
> Assuming you have a freshly wiped test cluster, with authorisation and repositories setup, but have previously setup a ceph.conf file, why do you
> once again have to run these commands in the correct order:
> 
> ceph-deploy install ceph-node2 ceph-node3 ceph-node4
> ceph-deploy mon create-initial
> ceph-deploy --overwrite-conf rgw create ceph-node3:onr
> 
> When all the data needed to define the commands is already available in ceph.conf.

Because ceph.conf doesn't describe a cluster.  It should be a minimal set 
of configs to bootstrap things (like, which mon addrs to try to get the 
*actual* state of the cluster, as maintained by the mons.)  It certainly 
wouldn't know anything about your rgw or mds or other instances.  We used 
to do that (defined all daemon in ceph.conf) but it's ungainly and 
difficult to maintain.

> If you had a model, that gathered the desired state, a way to establish the current state, a way to compare states, and a way to apply states,
> this could just be:
> 
> ceph-deploy apply-cfg
> 
> Essentially this is the strength of MVC. Also if you ran a redundant reapply ceph-deploy could simply gather the state and be able to tell you no
> operations are needed.
> 
> This is not hard to do, with the MVC design pattern.

...and if you want to take a cluster description and have a single button 
that makes it so, use chef or puppet -- that's what those tools are 
designed to do.

ceph-deploy is designed to do what you ask it to, and nothing more.  I am 
more likely to run a subset of the above commands than all of them in 
sequence.  e..g, I will upgrade with ceph-deploy install.  I will add a 
single osd with ceph-deploy osd create.  I will add a single rgw with 
ceph-deploy rgw create.  If you want to repeat the same sequence over 
again after wiping the cluster repeatedly (is this a test environment?)
then you can put it in a script to save typing...

osynge:
I think we both want things simple as possible, I assume you think the current model is simplest, and I think migration to MVC would make the current functionality simpler.

I think following a MVC pattern not only simpler, it is easier to maintain, easier to support cross platform operation, easier to add new functionality later easier, and easier to improve usability by allowing data defined orders of operation.

I realise that using an MVC pattern is more complex to understand in the abstract, than having a lot of functions which mange state conditionals and state transitions all in the same function and operate in simple loops.

I assert that MVC is more DRY, more testable, more extensible and clearer to read in practice.

osynge:
This is a community project, and it seems to me regardless of the validity of my arguments, we need to bring a community on board.

Sadly for my opinion, only one other member of the community outside my current employer has shown enthusiasm toward the MVC approach in private.

I feel it is wiser, in my own self interest to argue for other changes in ceph and ceph-deploy. Positions that are more accessible to be understood, and gain more credibility before arguing for such complex things as the use of MVC as a design pattern. I could not understand the full benefits, before I had transitioned a very very similar project to ceph-deploy, from a state similar to where ceph-deploy is now, to how I would like to see ceph-deploy, so it is not surprising that I find it hard to convince others.

I am not yet convinced it is worth gambling my time on proving my assertions that a migration toward MVC as a design pattern for ceph-deploy would show immediate benefits in:

 * testability.
 * minimising duplication.
 * maintenance.
 * extensible nature.

Of which I have no doubt would be achieved if I invest enough time. But time is limited and we all have more than one thing to do, especially in small teams like mine.

osynge:
On 08/05/2015 11:36 PM, Sage Weil wrote:
> On Wed, 5 Aug 2015, Owen Synge wrote:
>> Ok I try to be clearer:
>>
>> Assuming you have a freshly wiped test cluster, with authorisation and repositories setup, but have previously setup a ceph.conf file, why do you
>> once again have to run these commands in the correct order:
>>
>> ceph-deploy install ceph-node2 ceph-node3 ceph-node4
>> ceph-deploy mon create-initial
>> ceph-deploy --overwrite-conf rgw create ceph-node3:onr
>>
>> When all the data needed to define the commands is already available in ceph.conf.
> 
> Because ceph.conf doesn't describe a cluster.  It should be a minimal set 
> of configs to bootstrap things (like, which mon addrs to try to get the 
> *actual* state of the cluster, as maintained by the mons.)  It certainly 
> wouldn't know anything about your rgw or mds or other instances.  We used 
> to do that (defined all daemon in ceph.conf) but it's ungainly and 
> difficult to maintain.

For me at least it is irrelevant to the argument

If the state, is in the local config file or not in this particular case
I did not include the osd's or mention that my argument is valid for
osd's if the mons are already deployed, as state can be gathered then
from the mons.


>> If you had a model, that gathered the desired state, a way to establish the current state, a way to compare states, and a way to apply states,
>> this could just be:
>>
>> ceph-deploy apply-cfg
>>
>> Essentially this is the strength of MVC. Also if you ran a redundant reapply ceph-deploy could simply gather the state and be able to tell you no
>> operations are needed.
>>
>> This is not hard to do, with the MVC design pattern.
> 
> ...and if you want to take a cluster description and have a single button 
> that makes it so, use chef or puppet -- that's what those tools are 
> designed to do.

I am aware of this.

This is just an example, of one of the many things that falls out of an
MVC adoption with almost no work.

> ceph-deploy is designed to do what you ask it to, and nothing more.  I am 
> more likely to run a subset of the above commands than all of them in 
> sequence.  e..g, I will upgrade with ceph-deploy install.  I will add a 
> single osd with ceph-deploy osd create.  I will add a single rgw with 
> ceph-deploy rgw create.  If you want to repeat the same sequence over 
> again after wiping the cluster repeatedly (is this a test environment?)
> then you can put it in a script to save typing...

I agree, its just an example of the usability benefits, not expecting
the tester to write such scripts.

osynge:
Another clear example for MVC is once you have contacted the mons, and gathered the known OSD's, you know if you can activate them, before attempting to do so, you know if you have already created an OSD on a particular device before doing so.

This means you can present the user with much clearer error messages.

I wont go on about how many more complex things can be achieved, with further examples as of how separating discovering state, comparing state desired, to current cluster state, and applying state transitions based on the difference, would benefit ceph-deploy for end users, or maintainer developers.

osynge:
Oh and just one more obvious thing about MVC, 

Which I am sure you haver guessed and are aware of, with the MVC design pattern if done properly, is the ease of presenting what would be done if you ran a command. After all, doing some thing and presenting what would be done, are just alternative views :)

PR #324, opened 2015-07-10
Fix package installation on SUSE platforms v3

ddiss:
As raised:

Changes since v2 (#312):
- Remove unused imports (Thanks Travis)
- Rebase against latest upstream

Changes since v1 (#308):
- Test ceph-deploy component -> SUSE package name mapping (Thanks Ken)
- Rebase against latest upstream

### DIFF
--- a/ceph_deploy/hosts/suse/install.py
+++ b/ceph_deploy/hosts/suse/install.py
@@ -14,7 +14,7 @@ def install(distro, version_kind, version, adjust_repos, **kw):
 
 
 def mirror_install(distro, repo_url, gpg_url, adjust_repos, **kw):
-    # note: when split packages for ceph land for Suse,
+    # note: when split packages for ceph land for SUSE,
     # `kw['components']` will have those. Unused for now.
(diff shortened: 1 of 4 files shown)


trhoden:
Hmm, I wonder why Jenkins didn't test this PR...

ddiss:
looks like https://github.com/ceph/ceph-build/pull/97 is still open.

trhoden:
If that was it, it would ask for an admin to okay the test...

It looks like jenkins.ceph.com is down... :/

trhoden:
I was able to trigger the build manually: http://jenkins.ceph.com/job/ceph-deploy-pull-requests/1643/

The test failed because something weird with mock in Python 2.6 but the 2.7 tests pass and so does flake8, so I'm good with that.  :)

trhoden:
ok to test

trhoden:
lol, jenkins, you're catching up!

trhoden:
@ddiss Thanks David!

PR #309, opened 2015-06-12
Support iSCSI utility installation (targetcli)

ddiss:
This series adds two patches atop my previous pull-request at https://github.com/ceph/ceph-deploy/pull/308 to add support for iSCSI component installation, in preparation for clustered RBD/LIO iSCSI target deployment:
https://wiki.ceph.com/Planning/Blueprints/Hammer/Clustered_SCSI_target_using_RBD

PR #326, opened 2015-07-15
Support iSCSI utility installation (targetcli) v2

ddiss:
This series adds support for iSCSI component installation, in preparation for clustered RBD/LIO iSCSI target deployment:
https://wiki.ceph.com/Planning/Blueprints/Hammer/Clustered_SCSI_target_using_RBD

Changes since v1 (https://github.com/ceph/ceph-deploy/pull/309)
- Added test_install_only_iscsi unit test
- Rebased against master

### DIFF
--- a/ceph_deploy/hosts/suse/install.py
+++ b/ceph_deploy/hosts/suse/install.py
@@ -18,6 +18,8 @@ def map_components(components):
         packages.append('ceph-common')
     if 'ceph-radosgw' in components:
         packages.append('ceph-radosgw')
+    if 'targetcli' in components:
+        packages.append('targetcli')
 
     return packages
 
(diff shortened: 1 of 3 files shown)


trhoden:
@ddiss This seems to just use `--iscsi` as a shortcut to installing `targetcli`.

I think the main question is whether abstracting the ultimate package name away from the user is beneficial.  I'm hesitant to pull this in because `targetcli` is not a Ceph package, and `ceph-deploy` already has generic commands for installing packages by name.

You could do:

`ceph-deploy pkg --install targetcli <host>` for example.

ddiss:
I'd hoped to piggy-back on the "install --iscsi" behaviour to also perform sanity checks for the corresponding Ceph RBD and LIO kernel modules. However, this could be done outside of ceph-deploy, and I agree that the "ceph-deploy pkg --install" functionality should be sufficient.

Thanks for the feedback Travis.

PR #329, opened 2015-07-17
Class-based package managers to add DNF support for Fedora 22

trhoden:
This PR does two things:

(1) uses DNF as the package manager for Fedora >= 22
(2) Introduces Yum and DNF package manager classes

The new package manager classes can be instantiated at the time of a call to hosts.get(), with the module only needing to decide once which manager is the correct one.  I've only added this to Fedora right now, as an early adopter of the new code.

I would like to use this everywhere, and think it makes the code a lot cleaner.

Specifically for this change, Fedora installs are a bit weird because F22 packages are only available from the gitbuilders, so you have to do "ceph-deploy install --dev master <HOST>" to try it out.  Becuase it's a dev install, it actually uses centos.mirror_install() :disappointed: to write out a .repo file, so you will still see some references to Yum until  that is abstracted out in a different way.

The other thing that was not clear to me for DNF is that right now we require the Yum priorities plugin.  DNF supports repo priorities natively, so we don't need that with DNF (I have that).  However, we *also* need to configure the priorities plugin with `check_obsoletes=1`, and I couldn't find anything about if there is an equivalent setting in DNF (what it is, where it would be set).

http://tracker.ceph.com/issues/12259

### DIFF
--- a/ceph_deploy/util/pkg_managers.py
+++ b/ceph_deploy/util/pkg_managers.py
@@ -177,3 +177,102 @@ def zypper_refresh(conn):
         conn,
         cmd
     )
+
+
+class PackageManager(object):
+    """
+    Base class for all Package Managers
+    """
+
+    def __init__(self, remote_conn):
+        self.remote_info = remote_conn
+        self.remote_conn = remote_conn.conn
+
+    def _run(self, cmd):
+        return remoto.process.run(
+            self.remote_conn,
+            cmd
+        )
+
+    def install(self, packages):
+        """Install packages on remote node"""
+        raise NotImplementedError()
+
+    def remove(self, packages):
+        """Uninstall packages on remote node"""
+        raise NotImplementedError()
+
+    def clean(self):
+        """Clean metadata/cache"""
+        raise NotImplementedError()
+
+
+class RPMManagerBase(PackageManager):
+    """
+    Base class to hold common pieces of Yum and DNF
+    """
+
+    executable = None
+
+    def install(self, packages):
+        if isinstance(packages, str):
+            packages = [packages]
+
+        cmd = [
+            self.executable,
+            '-y',
+            'install',
+        ]
+        cmd.extend(packages)
+        return self._run(cmd)
+
+    def remove(self, packages):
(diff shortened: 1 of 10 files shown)


ceph_deploy/util/pkg_managers.py
```diff
@@ -177,3 +177,93 @@ def zypper_refresh(conn):
...
+        return self._run(cmd)
+
+
+class DNF(RPMManagerBase):
+    """
+    The DNF Package manager
+    """
+
+    def __init__(self, remote_conn):
+        super(DNF, self).__init__(remote_conn, 'dnf')
```
alfredodeza: I don't see the need to require passing `cmd` to `__init__` because the classes use `cmd` as well that is different from `self.cmd` and because this can just be defined as a class attribute:

    class DNF(RPMManagerBase):

        executable = 'dnf'

It doesn't need to be named `executable` but the idea is that this would skip the need to requiring it

alfredodeza: Actually, this would prevent the need for using `super` also, the base classes don't need to call it at all  (unless I am reading this wrong)

trhoden: @alfredodeza 
It would eliminate the need for `super`, but I don't quite see how it would work if `self.cmd` does not get set in `RPMManagerBase`.

If I do `Yum().install('tree')`, The `Yum` class does not override the `install` method -- it falls through to `RPMManagerBase`.    If `Yum` does not pass the executable name to `RPMManagerBase` via `__init__()`, how will it know which executable to use?

I don't think a class attribute would do that for me.

On a side note, I would be happy to rename `RPMManagerBase.cmd` to `executable` so it does not get confused with the `cmd` string I pass to `_run()`.

trhoden:
@dmick @alfredodeza 

At a high level, though, do you find this a useful pursuit?

dmick:
I certainly do, as I was just bitching about ansible not providing generic package services yesterday :)

The abstraction will never be perfect, but I say let's not assassinate the good for that reason alone.

alfredodeza:
+1

PR #330, opened 2015-07-23
More help/argparse cleanup

trhoden:
Most of what I had intended to under http://tracker.ceph.com/issues/12083 got split out into different issues and have already been resolved (Yay!).

This is just a couple of minor things that I had queued up that didn't go in yet.

The `prog` variable in the argparse namespace always confused me.  It's called out as an "ugly kludge", but it's not actually used anywhere.  Kill it!

### DIFF
--- a/ceph_deploy/admin.py
+++ b/ceph_deploy/admin.py
@@ -59,7 +59,7 @@ def make(parser):
         'client',
         metavar='HOST',
         nargs='+',
-        help='host to configure for ceph administration',
+        help='host to configure for Ceph administration',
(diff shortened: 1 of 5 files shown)


dmick:
I think the point of prog had been to pass the name of the program down to the subparsers for error messages.  I imagine it went away when the logging got all Technicolor.

trhoden:
Makes sense, Dan.  I didn't go far enough back into history to see what it was ever used for, but I don't think it's been used in at least a year. :)

PR #332, opened 2015-07-31
systemd changes ... it's working on fc22!

liewegas:


ceph_deploy/hosts/fedora/__init__.py
```diff
@@ -18,7 +18,7 @@ def choose_init():
 
     Returns the name of a init system (upstart, sysvinit ...).
     """
-    return 'sysvinit'
+    return 'systemd'
```
trhoden: Do we really want to switch this over wholesale?

F20 was EOL'd on 6/23, but we do still have packages for F20 in rpm-hammer.  We never built F21 packages.

My concern is that if we took this is in as-is, and I push a newer version of ceph-deploy into rpm-hammer, then ceph-deploy won't be able to work with the F20 packages that are also present in rpm-hammer.

I'm not sure if this is an issue or not.

liewegas: I'm less worried about fc21...

The bigger problem is that with these changes you can no longer install hammer, which is the currently recommended stable release.  We need to figure out how to make the init system choice dependent on the Ceph version installed...

trhoden:
Other than the question about cutting off support for F20, this all looks good to me.

I think the whole thing will get re-worked quite a bit once we switch to a more abstracted view of init systems, but it looks good right now.

liewegas:
I tried out centos7 today and ended up having to duplicate all of fedora/mon/create.py to centos7/mon/create.py (and will have to do the same with rhel).  If you have even a half-baked plan on avoiding all that duplication I say let's do it...

ceph_deploy/hosts/fedora/__init__.py
trhoden: Would you be open to at least making it a clear cut-over with F22, like it's done with Yum-> DNF?

https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/hosts/fedora/__init__.py#L25-L28

But yeah, I see your point about versions.  That will take a bit more thought.

Are you picturing wanting to release systemd-enabled packages on F22?  Unless such packages are going to show up, it's not an issue for Hammer. 

But for Infernalis (maybe?) and definitely Jewel, we will have packages that are now systemd where we were sysvinit before (-EL7 anything).  So yeah, we'll have to be smart enough for that.   :/

This is a problem that has to be solved anyways, because we will also need to make different choice when it comes to osd/mon package split depending on Ceph version.

I think a simple table can be used for named releases. but for dev/gitbuilder builds, that's a bit uglier.  Then we ahve to start hardcoding entries of "if version >= 9.x....".  We'll see.

trhoden:
Yes, there is definite a bit more than a half-baked plan.  It may be about a week before I can get to it, though.

My basic idea is make the init system handler be a class that gets attached when the module is chosen, just like I've started doing for the package manager.  See #334.

ceph_deploy/hosts/fedora/__init__.py
liewegas: Making a conditional yum/dnf switchover seems fine.

The hope is to get these systemd changes merged this week for infernalis, which means that any community users will need the ceph-deploy smarts to handle sysvinit vs systemd.  :/

liewegas:
Sounds good.  Timeframe sounds fine too as long as we'll be able to make fedora+centos+rhel work  before infernalis gets an RC.

ceph_deploy/hosts/fedora/__init__.py
trhoden: Just to follow up - I was asking if we could add that conditional to this PR.

But, I'm happy to merge this in and then add the conditional on top.

I think the likely way forward for this is that this set of changes will go into a 1.5.27 version, and then we can focus on sweeping systemd changes for 1.5.28.

liewegas:
Added a patch that makes fedora conditional on fc22 or later

liewegas:
Ok, dropped that comment (see wip-systemd-fchack) since my python is terrible and broken.

trhoden:
Going to merge this in.  Will go into 1.5.27, then we will use it as a basis when refactoring a common approach to systemd and other init systems for 1.5.28.

PR #333, opened 2015-07-31
Only install requested packages for Debian, CentOS, Fedora

trhoden:
http://tracker.ceph.com/issues/12543

This PR makes it such that `install --rgw` only installs radosgw, `--cli` only installs ceph-common, etc.

This does preserve backwards-compatibility. The default is still "everything", where everything is MON, OSD, MDS, and RGW.  This just makes it such that if a user requests specific daemons, we install the minimal set of packages that can satisfy the request.

That set of packages can be different depending on OS.  For example, debian packages have a separate `ceph-mds` package (enabling it to be installed by itself), but EL/RPM packages do not.

I tested that this didn't break anything by running this sequence of commands on Trusty, CentOS 7, and Fedora 22:

```
ceph-deploy install --cli <node>
ceph-deploy install --rgw <node>
ceph-deploy install --mon --osd <node>
ceph-deploy install --mds <node>
ceph-deploy install <node>
ceph-deploy install --dev master <node>
ceph-deploy install --dev master --rgw <node>
```

Except for on Fedora, I could only test with ``--dev master`` since there are only F22 packages on gitbuilder.

In each case, I looked at the top-level packages listed by Yum/Apt/DNF to see it was the expected set.  All looked good!

### DIFF
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -1,7 +1,11 @@
 from ceph_deploy.util import pkg_managers, templates
 from ceph_deploy.lib import remoto
 from ceph_deploy.hosts.util import install_yum_priorities
+from ceph_deploy.hosts.common import map_components
 from ceph_deploy.util.paths import gpg
+
+
+NON_SPLIT_PACKAGES = ['ceph-osd', 'ceph-mon', 'ceph-mds']
 
 
 def rpm_dist(distro):
@@ -36,8 +40,10 @@ def repository_url_part(distro):
 
 
 def install(distro, version_kind, version, adjust_repos, **kw):
-    # note: when split packages for ceph land for CentOS, `kw['components']`
-    # will have those. Unused for now.
+    packages = map_components(
+        NON_SPLIT_PACKAGES,
+        kw.pop('components', [])
+    )
     logger = distro.conn.logger
     release = distro.release
     machine = distro.machine_type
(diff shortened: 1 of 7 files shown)


ceph_deploy/hosts/centos/install.py
```diff
@@ -4,6 +4,23 @@
 from ceph_deploy.util.paths import gpg
 
 
+def map_components(components):
```
alfredodeza: this is almost the same in the other distros, except for the packages that are split differently.

Would it be better if something does the work of intersecting the lists to determine the packages needed? It seems that
they all follow the same pattern, except for the change of items that pull in the ceph package. So something like:

    packages = set()
    for c in ceph_components:
        if c in components:
            components.remove(c)
            packages.append('ceph')

    for c in components:
        packages.append(c)

    return list(packages)

Would allow us to just define what the `ceph_components` are (mind you I didn't put much thought in naming) and the helper would get the same results.

alfredodeza: it would allow you to unittest it  too!

trhoden: alright, I follow...  So the only thing that would be distro specific would be set of components that map to 'ceph' instead of their original name...

That makes sense.  :)  I'll give it a shot.

alfredodeza: sorry, maybe some more context would've helped. Don't make this tiny comment prevent you from moving forward though!

trhoden:
@alfredodeza See if you like this better.  :)

PR #334, opened 2015-08-03
Use package manager classes for remaining distros

trhoden:
First added for Fedora/DNF, move all the other install/uninstall tasks to using the pkg manager classes.

There is still more work to do on top of this, but this is a big start.

Tested install/uninstall on CentOS 7, Trusty, and RHEL 7 and had no problems here.

http://tracker.ceph.com/issues/12439

### DIFF
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -1,6 +1,5 @@
-from ceph_deploy.util import pkg_managers, templates
+from ceph_deploy.util import templates
 from ceph_deploy.lib import remoto
-from ceph_deploy.hosts.util import install_yum_priorities
 from ceph_deploy.hosts.common import map_components
 from ceph_deploy.util.paths import gpg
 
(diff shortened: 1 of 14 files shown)


trhoden:
This has been re-based after the merge of #333, and should be good to go know.

@alfredodeza mind giving it a look?

ceph_deploy/hosts/centos/install.py
```diff
@@ -161,13 +141,13 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True,
...
         # set the right priority
-        install_yum_priorities(distro)
+        distro.packager.install_priorities_plugin()
         distro.conn.remote_module.set_repo_priority(['Ceph', 'Ceph-noarch', 'ceph-source'])
         distro.conn.logger.warning('alter.d ceph.repo priorities to contain: priority=1')
 
 
-    if extra_installs:
-        pkg_managers.yum(distro.conn, packages)
+    if extra_installs and len(packages):
```
alfredodeza: why `len(packages)`? lists evaluate to `False` when empty and `True` when not, so you could do `if extra_installs and pacckages:`

trhoden:
@alfredodeza extraneous calls to `len()` removed.

ceph_deploy/hosts/suse/install.py
```diff
@@ -15,9 +15,9 @@ def install(distro, version_kind, version, adjust_repos, **kw):
         kw.get('components', [])
     )
 
-    pkg_managers.zypper_refresh(distro.conn)
-    if len(packages):
```
alfredodeza: this was probably me right? :(

ceph_deploy/hosts/suse/install.py
```diff
@@ -92,5 +92,5 @@ def repo_install(distro, reponame, baseurl, gpgkey, **kw):
     )
 
     # Some custom repos do not need to install ceph
-    if install_ceph and len(packages):
-        pkg_managers.zypper(distro.conn, packages)
+    if install_ceph and packages:
```
alfredodeza: this looks beautiful

PR #335, opened 2015-08-04
Add "ceph-deploy repo" command

trhoden:
This introduces a new `ceph-deploy repo` set of commands.

You can use `ceph-deploy repo` to create or delete repo files by name.  The repo details can be passed in my `--gpg-url` and `--repo-url`.  Note that for RPM systems, this command only supports the basic "ceph" template -- meaning that it will always put three entries into a single `.repo` file for [ceph], [ceph-noarch], and [ceph-source].

We could potentially swap this out for the more flexible custom/arbitrary repo definitions if that is deemed worthwhile.  Even if we think that it might be, let's talk about it because I don't want to **not** be able to switch to it later for backwards compatibility reasons because we didn't start that way.

I haven't tested yet that this works when reading repos from `cephdeploy.conf`.  I need to check that next.  I did test that `--remove` worked on CentOS7 and Trusty, and that I could create valid repo definitions using `--gpg-url` and `--repo-url` on those same two distros.

Another future TODO is to add support for reading the repo/gpg URL from env vars (as is done in another place of the code) so that security conscious users can use those to input URLs that use HTTP basic auth (e.g. https://{user}:{password}@fqdn/repo/....).

### DIFF
--- a/ceph_deploy/repo.py
+++ b/ceph_deploy/repo.py
@@ -1,0 +1,50 @@
+import logging
+
+from ceph_deploy.cliutil import priority
+
+
+LOG = logging.getLogger(__name__)
+
+
+def repo(args):
+    pass
+
+
+@priority(70)
+def make(parser):
+    """
+    Repo definition management
+    """
+
+    parser.add_argument(
+        'repo_name',
+        metavar='REPO-NAME',
+        help='Name of repo to manage.  Can match an entry in cephdeploy.conf'
+    )
+
+    parser.add_argument(
+        '--repo-url',
+        help='a repo URL that mirrors/contains Ceph packages'
+    )
+
+    parser.add_argument(
+        '--gpg-url',
+        help='a GPG key URL to be used with custom repos'
+    )
+
+    parser.add_argument(
+        '--remove', '--delete',
+        action='store_true',
+        help='remove repo definition on remote host'
+    )
+
+    parser.add_argument(
+        'host',
+        metavar='HOST',
+        nargs='+',
+        help='host(s) to install on'
+    )
+
+    parser.set_defaults(
+        func=repo
+    )
--- a/ceph_deploy/repo.py
+++ b/ceph_deploy/repo.py
@@ -7,6 +7,47 @@ from ceph_deploy.cliutil import priority
 LOG = logging.getLogger(__name__)
 
 
+def install_repo(distro, args, cd_conf, rlogger):
+    if args.repo_name in cd_conf.get_repos():
+        LOG.info('will use repository %s from ceph-deploy config', args.repo_name)
+        options = dict(cd_conf.items(args.repo_name))
+        extra_repos = cd_conf.get_list(args.repo_name, 'extra-repos')
+        try:
+            repo_url = options.pop('baseurl')
+            gpg_url = options.pop('gpgkey', None)
+        except KeyError as err:
+            raise RuntimeError(
+                'missing required key: %s in config section: %s' % (err, args.repo_name)
+            )
+    else:
+        repo_url = args.repo_url
+        gpg_url = args.gpg_url
+        extra_repos = []
+
+    repo_url = repo_url.strip('/')  # Remove trailing slashes
+    distro.packager.add_repo(
+        args.repo_name,
+        repo_url,
+        gpg_url=gpg_url
+    )
+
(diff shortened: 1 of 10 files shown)


ceph_deploy/repo.py
```diff
@@ -0,0 +1,112 @@
...
+    if args.repo_name in cd_conf.get_repos():
+        LOG.info('will use repository %s from ceph-deploy config', args.repo_name)
+        options = dict(cd_conf.items(args.repo_name))
+        extra_repos = cd_conf.get_list(args.repo_name, 'extra-repos')
+        try:
+            repo_url = options.pop('baseurl')
+            gpg_url = options.pop('gpgkey')
+        except KeyError as err:
+            raise RuntimeError(
+                'missing required key: %s in config section: %s' % (err, args.repo_name)
```
alfredodeza: it is awesome that you are taking the time to properly return meaningful error responses. Thank you.

ceph_deploy/util/pkg_managers.py
```diff
@@ -206,6 +210,18 @@ def clean(self):
         """Clean metadata/cache"""
         raise NotImplementedError()
 
+    def add_repo_gpg_key(self, url):
+        """Add given GPG key for repo verification"""
+        raise NotImplementedError()
+
+    def add_repo(self, name, url, gpg_url):
```
alfredodeza: to avoid the problem of future backward incompatibility I would just slap a `**kw` here and pass in name, url, gpg_url, and anything else as keyword arguments. That way you can use what you are being passed in today, but you are allowing passing more later without having to change the signature (and possibly making it look odd)

ceph_deploy/util/templates.py
```diff
@@ -1,7 +1,6 @@
 
 
-ceph_repo = """
-[ceph]
```
alfredodeza: good catch

ceph_deploy/repo.py
trhoden: Thank yourself!  It's a direct lift from: https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/install.py#L246

I'm hopeful to eventually be able to make the `install --repo` code use this `repo` code directly.

ceph_deploy/util/pkg_managers.py
trhoden: Are you saying to handle the existing positional args args inside `**kw` instead?

Adding `**kw` seems like a good idea.  My personal preference is to make *required* args be clear/obvious by making them positional.  But it is just that -- a personal preference.  So I would lobby for adding `**kw` and leaving the existing ones.  What say you?  :)

ceph_deploy/util/templates.py
trhoden: The suse/zypper one had already been changed, and I looked at it and said 'why is that one funny?'.  Then I ran the code and it was a 'duh' moment.  :)

ktdreyer:
The split between "ceph" and "ceph-noarch" is something that happened log ago and there's no technical reason to maintain such a split today. Ideally we would ship all noarch packages into the respective binary locations (ie, copy all noarch packages to `/x86_64/` and `/i686/` and `/arm/`, etc.)

Removing "ceph-noarch" is going to require some non-trivial buildsystem changes and it's really far down on my todo list. But eventually we should remove it to make it easier for users.

You'd mentioned backwards-compatibility and that's the only thing that came to my mind here.

I guess once we drop "ceph-noarch" in the build system, we'll just remove it from ceph-deploy too.

trhoden:
@ktdreyer Thanks for that feedback.  It would be a lot nicer if ceph-noarch was not in the picture, and I didn't really understand it's purpose.

Perhaps in the shorter term, we can copy the noarch packages into $basearch?

Either way, i think I have a way forward for now.

trhoden:
Major updates since last review:

* allow repo and gpg URLS to be passed by env var
* create APT repo file with mode 0600 if URL contains a password (HTTP basic auth)
* --gpg-url no longer required.  if present, set yum gpgcheck=1, set =0 if not
* Add Apt package pinning post repo creation

Now I need to test this on EL 7 and Trusty.  Need to make sure it works with ice-setup, and that referencing repos by name from cephdeploy.conf actually works.

trhoden:
@alfredodeza All my tests looked good.  :+1: 

Do with this what you will.

PR #336, opened 2015-08-04
Only install priorities plugin when using Yum

trhoden:
Since DNF natively supports priorities, just skip all the priorities stuff if we're not Yum.

Skip the check_obsoletes on DNF as well.

I didn't like having a "install_specific_package...()" method in the package manager classes, so I created a helper to return the name and just used the regular install() method instead.

Tested on CentOS and Fedora.  Seems to work!

http://tracker.ceph.com/issues/12477

### DIFF
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -54,7 +54,7 @@ def install(distro, version_kind, version, adjust_repos, **kw):
     # Get EPEL installed before we continue:
     if adjust_repos:
         distro.packager.install('epel-release')
-        distro.packager.install_priorities_plugin()
+        distro.packager.install('yum-plugin-priorities')
         distro.conn.remote_module.enable_yum_priority_obsoletes()
         logger.warning('check_obsoletes has been enabled for Yum priorities plugin')
     if version_kind in ['stable', 'testing']:
@@ -141,9 +141,10 @@ def mirror_install(distro, repo_url, gpg_url, adjust_repos, extra_installs=True,
 
         distro.conn.remote_module.write_yum_repo(ceph_repo_content)
         # set the right priority
-        distro.packager.install_priorities_plugin()
+        if distro.packager.name == 'yum':
+            distro.packager.install('yum-plugin-priorities')
         distro.conn.remote_module.set_repo_priority(['Ceph', 'Ceph-noarch', 'ceph-source'])
-        distro.conn.logger.warning('alter.d ceph.repo priorities to contain: priority=1')
+        distro.conn.logger.warning('altered ceph.repo priorities to contain: priority=1')
 
 
     if extra_installs and packages:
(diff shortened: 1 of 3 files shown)


ceph_deploy/hosts/centos/install.py
```diff
@@ -198,11 +200,8 @@ def repo_install(distro, reponame, baseurl, gpgkey, **kw):
 
     # set the right priority
     if kw.get('priority'):
-        distro.packager.install_priorities_plugin()
-        logger.warning(
-            'ensuring that {repo_path} contains a high priority'.format(
-                repo_path=repo_path)
-        )
+        if isinstance(distro.packager, Yum):
```
alfredodeza: it seems that it would be good to have a way to tell if this is yum besides using `isinstance`. I know that we have `executable` and you could do: `if distro.packager.executable == 'yum':` but it seems that for Apt we are using a list.

I don't think that a list is needed in `Apt` and could be normalized to `apt-get` and then this would work.

Otherwise, you may want to add something like `.name` ? 

Either way, I see this being used often and might be a good idea.

trhoden: I'm surprised - do you consider `isinstance` to be bad practice?  It seems like a very OO way of doing things.  We don't need to know anything about the implementation of the class (i.e. attributes like `name` or `executable`, just merely that the packager is a class doing Yum.

I also like that if I didn't care whether it was Yum or DNF, I could do `isinstance(distro.packager, RPMBaseManager)`.

What are the drawbacks of `isinstance` ?

> I don't think that a list is needed in Apt and could be normalized to apt-get and then this would work.

Whether to switch on `executable` aside, I'm happy to make `executable` be just a single executable name (`apt-get` for Apt) and add the common options another way.  I was just getting lazy. :)

trhoden:
I just push an update that fixes the CentOS version to priorities plugin name mapping...  It never looked right to me, but I had just cut/paste what was elsewhere.

It's `yum-priorities` on CentOS 5 (not 6), and `yum-plugin-priorities` from then on.

Given that we don't even have EL 5 packages in rpm-firefly, that makes me thing we could just drop the mapping and always install `yum-plugin-priorities` entirely.  Thoughts?

liewegas:
We can definitely ignore el5.  :)

trhoden:
@liewegas out-of-date reference `yum-priorities` dropped. :)

ceph_deploy/hosts/centos/install.py
alfredodeza: I didn't mean it was bad practice, if you don't foresee checking the name of the package manager often then this would work fine.

If this is not a one-off, then you are forcing an import of the class to check. Since this is a class you control, adding an attribute to help these types of checks would be nice.

trhoden: It's definitely not a one-off. I'll add a `name` attribute.

PR #337, opened 2015-08-05
Allow package upgrades with DNF

trhoden:
Doing `dnf install pkg` does not actually upgrade the package if it is already installed.  But adding `--best` will make it do so.

There are more commits here than necessary to just add `--test` to the DNF call.  The other commits are to modify the install() and remove() methods of the package manager classes to take `**kw`, and to pass keyword args for adding extra flags to install and remove calls.

The individual commits make sense.

http://tracker.ceph.com/issues/12480

### DIFF
--- a/ceph_deploy/util/pkg_managers.py
+++ b/ceph_deploy/util/pkg_managers.py
@@ -199,7 +199,7 @@ class PackageManager(object):
             **kw
         )
 
-    def install(self, packages):
+    def install(self, packages, **kw):
         """Install packages on remote node"""
         raise NotImplementedError()
 
@@ -232,15 +232,21 @@ class RPMManagerBase(PackageManager):
     executable = None
     name = None
 
-    def install(self, packages):
+    def install(self, packages, **kw):
         if isinstance(packages, str):
             packages = [packages]
 
+        extra_flags = kw.pop('extra_install_flags', None)
         cmd = [
             self.executable,
             '-y',
             'install',
         ]
+        if extra_flags:
+            if isinstance(extra_flags, str):
+                extra_flags = [extra_flags]
+            cmd.extend(extra_flags)
+
         cmd.extend(packages)
         return self._run(cmd)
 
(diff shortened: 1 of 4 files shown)


trhoden:
@liewegas adding `--best` to DNF seems to solve the DNF not-upgrading issue.

We've never supported downgrading that I know of...

trhoden:
@alfredodeza this is ready for review.

I tested install/uninstall/purge on Trusty to make sure that still worked.

On F22, I did install/uninstall/install/upgrade.  All worked.

ceph_deploy/util/pkg_managers.py
```diff
@@ -319,6 +330,16 @@ class DNF(RPMManagerBase):
     executable = 'dnf'
     name = 'dnf'
 
+    def install(self, packages, **kw):
+        extra_install_flags = kw.pop('extra_install_flags', [])
+        if '--best' not in extra_install_flags:
```
alfredodeza: will there never be a situation where we do not want `--best` ? (because it looks like we are forcing it regardless)

trhoden: You are right that we are forcing it...

I'm not sure to tell you the truth.  I am making it behave like Yum and apt-get, such that installing a package that is already present will upgrade it.  DNF just says "it's already there".  So it's consistent in that regard.  Let me do one more check to make sure I am right about this on Yum...

trhoden: So yes, Yum automatically updates packages, and DNF does not.

Doing the following on CentOS 7:

```
ceph-deploy install --dev next cent701
ceph-deploy install --dev master cent701
```

does an upgrade, even though the underlying commands are just `yum -y install ceph ceph-radosgw` in each case.

Adding `--best` seems to recreate this behavior.  Of course DNF is new and there may be consequences to `--best` we do not understand yet.  The docs mention that you are more likely to get into dependency resolution problems because you are telling it to always use the newer version of the package, even if it can't resolve the deps.

If that happens, that probably means there's a problem with the packages, IMO.  We'll see!

PR #339, opened 2015-08-06
Do not enable non-existing ceph service on systemd managed EL distros.

mbroz:
Once systemd integration is ready, "systemctl enable ceph" fails
because there is no longer such service (OSD units are generated).

Also see https://github.com/ceph/ceph-deploy/commit/8360fb501aa8a3b8dc20423c1575379b3d30690e

### DIFF
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -379,8 +379,10 @@ def activate(args, cfg):
         time.sleep(5)
         catch_osd_errors(distro.conn, distro.conn.logger, args)
 
-        if distro.is_el:
-            system.enable_service(distro.conn)
+        if distro.init == 'systemd':
+            system.enable_service(distro.conn, "ceph.target")
+        elif distro.init == 'sysvinit':
+            system.enable_service(distro.conn, "ceph")
 
         distro.conn.exit()
 
--- a/ceph_deploy/hosts/__init__.py
+++ b/ceph_deploy/hosts/__init__.py
@@ -68,7 +68,7 @@ def get(hostname,
     module.codename = codename
     module.conn = conn
     module.machine_type = machine_type
-    module.init = module.choose_init()
+    module.init = module.choose_init(module)
     module.packager = module.get_packager(module)
     return module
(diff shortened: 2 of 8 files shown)


ceph_deploy/osd.py
```diff
@@ -379,7 +379,7 @@ def activate(args, cfg):
         time.sleep(5)
         catch_osd_errors(distro.conn, distro.conn.logger, args)
 
-        if distro.is_el:
+        if distro.is_el and distro.init != 'systemd':
             system.enable_service(distro.conn)
```
liewegas: oh, can you make this an else enable ceph.target ?  this was one of the last things on my list for the basic fc22 support

liewegas: or rather, i think it should be if systemd (enable ceph.target) else if sysvinit (enable ceph) .. it has nothing to do with is_el, right?

mbroz:
@liewegas I think ceph.target is already enabled in code before it, at least I see
[ceph-node3][INFO  ] Running command: systemctl enable ceph.target
in log...

TBH I am not sure what is the real reason for if distro.is_el: ... shouldn't be ther one common if for systemd/sysvinit/upstart instead?

mbroz:
I'll change it to if (systemd) else if(sysvinit).... the is_el option do not make sense anymore,both Fedora na RHEL7 derivates will use systemd-enabled ceph soon...

mbroz:
Patch updated...

trhoden:
@mbroz This looks great.  Thanks!

I'm going to redo a lot of things for the systemd changes, and this is perfectly in step with the direction it's heading.

Ultimately this will end up such that instead of:

```
if distro.init == 'systemd':	
    system.enable_service(distro.conn, "ceph.target")
elif distro.init == 'sysvinit':
    system.enable_service(distro.conn, "ceph")
```

It will become:

```
distro.init.enable('ceph')
```

And `distro.init` will already be an instance of an Init class that is either systemd/sysvinit/upstart that will "do the right thing".

PR #344, opened 2015-08-11
Detect and support both systemd and sysvinit on Fedora, RHEL and CentOS

mbroz:
To properly install systemd-enabled and old (sysvinit) Ceph packages
we need to properly detect when systemd target is used.

This cannot be done just by checking distro version, it depends
on Ceph package content (sysvinit can be used on systemd distro as well).

Let's check ceph.target existence and enable systemd only in this case.

### DIFF
--- a/ceph_deploy/hosts/centos/__init__.py
+++ b/ceph_deploy/hosts/centos/__init__.py
@@ -17,8 +17,14 @@ def choose_init(module):
 
     Returns the name of a init system (upstart, sysvinit ...).
     """
-    return 'sysvinit'
 
+    if module.normalized_release.int_major < 7:
+        return 'sysvinit'
+
+    if not module.conn.remote_module.path_exists("/usr/lib/systemd/system/ceph.target"):
+        return 'sysvinit'
+
+    return 'systemd'
 
 def get_packager(module):
     return pkg_managers.Yum(module)
--- a/ceph_deploy/hosts/centos/mon/create.py
+++ b/ceph_deploy/hosts/centos/mon/create.py
@@ -6,19 +6,51 @@ from ceph_deploy.lib import remoto
 def create(distro, args, monitor_keyring):
     hostname = distro.conn.remote_module.shortname()
     common.mon_create(distro, args, monitor_keyring, hostname)
-    service = distro.conn.remote_module.which_service()
 
-    remoto.process.run(
-        distro.conn,
-        [
-            service,
-            'ceph',
-            '-c',
-            '/etc/ceph/{cluster}.conf'.format(cluster=args.cluster),
-            'start',
-            'mon.{hostname}'.format(hostname=hostname)
-        ],
-        timeout=7,
-    )
+    if distro.init == 'sysvinit':
+        service = distro.conn.remote_module.which_service()
+        remoto.process.run(
+            distro.conn,
+            [
+                service,
+                'ceph',
+                '-c',
+                '/etc/ceph/{cluster}.conf'.format(cluster=args.cluster),
+                'start',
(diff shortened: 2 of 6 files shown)


trhoden:
I had no idea you were working on this!

Both @alfredodeza and I had been working on this as well, but with greater ambitions of using a class-based init system manager.

In the interest of time and testing, it seems like this is a good fit and we can merge it in, though it is unlikely to remain in it's current form.  But we have other things to work on for now, and I'm happy to see this go in and be working.

My main concern is keying things off of the presence of the `ceph.target` file.  Do you think this is reliable enough?  There is discussion elsewhere of systemd being a runtime option on Debian distros (which this PR does not address), to the effect that it could be possible that the ceph.target file is present but a user has chosen not to use systemd.  I know that's a bigger question than you can answer, because it has a lot to do with how we formulate the ceph packages...

cc/ @liewegas @ktdreyer

mbroz:
Well, I am switching between systemd/sysvinit branches and ceph-deploy is without this big pain so I spent some time on this (a lot of more than expected though...)
I tried some more universal approaches and TBH this is the only one which works reliably for me (I tried "systemctl status ceph.target" to avoid path, for some reason this did not worked there).

@trhoden I think that systemd location and target name is stable enough here (even for Debian). But if there is anything better let's use that...

alfredodeza:
this looks ok to me, I think the things I was working on can still make it in at some point without dropping this effort.

liewegas:
The preferred strategy for debian-based systems is to install both systemd and sysvinit (and upstart) files, since the init system is a runtime option and not dictated by the os version.  That means keying off of ceph.target won't work there.  (But those packags aren't even built yet, so this is an okay interim solution.. as long as we do a proper check for what pid 1 is later.)

I think the right way is to look at pid 1 and see what it is.  For example, see  https://wiki.ubuntu.com/SystemdForUpstartUsers, which tells you what binary names to look for in pid 1.

trhoden:
In that case, then I think it because a combination -- see if systemd is running *and* check if ceph.target exists.  That should make us cover all cases, I think.

We do already have this: https://github.com/ceph/ceph-deploy/blob/d7987c20b526c3c6862d84ed5e075befdeddf5ce/ceph_deploy/util/system.py#L19-L31 for seeing if systemd is running.  Hopefully that can be reused, and if not, enhanced.

trhoden:
I guess a follow-up question is, does the strategy in this PR work for -EL distros?  We are talking about how the simple ceph.target detection won't work for debian, but does work for Fedora, CentOS, and RHEL?

Is systemd a runtime option on those distros?

mbroz:
@trhoden I tried some scratch builds for Fedora and CentOS and it worked (both systemd and old sysvinit). RHEL should be the same. And systemd is the only option, I think most packages already removed old sysvinit script completely. (read: for all supported Fedoras and RHEL7+ there is only one init system - systemd). So the question is only if ceph package supports systemd natively or it uses sysvinit compatibility handling.

mbroz:
And TBH this patch intentionally says Fedora/Centos/RHEL, I did not want to touch Debian/Ubuntu/Suse path, someone else should fix it there, probably with additional runtime check mentioned above...

trhoden:
@mbroz Apologies that I had steered the discussion towards debian.  My most recent comment was more of an effort to ask "hey, regardless of what we do with debain, it looks like this is probably good for -EL distros".  Meaning this can probably be merged as-is.

liewegas:
yep!

liewegas:
@trhoden Verified that the existing is_systemd() method will work on both fedora and centos (/proc/1/comm is "systemd").

PR #345, opened 2015-08-12
check if ceph is installed

alfredodeza:
This is a PR with a bit of feature creep: the intent was to use the new `Ceph` object to provide an interface for versions and use it to map it against supported `systemd` versions.

It is not entirely sure if this approach will be used, but for now this can be used to fix (http://tracker.ceph.com/issues/11115) which allows to check if Ceph is installed and error accordingly. 

The "check if ceph is installed" comes free with this implementation, if we decide later to not use versions we can change the approach.

This PR also moves the private function in `hosts/__init__.py` to `util.versions` because the functionality is now being used by more than one place. At some point the tests should map to the new location - but all tests pass as it is now.

### DIFF
--- a/ceph_deploy/hosts/__init__.py
+++ b/ceph_deploy/hosts/__init__.py
@@ -17,7 +17,8 @@ def get(hostname,
         username=None,
         fallback=None,
         detect_sudo=True,
-        use_rhceph=False):
+        use_rhceph=False,
+        callbacks=None):
     """
     Retrieve the module that matches the distribution of a ``hostname``. This
     function will connect to that host and retrieve the distribution
@@ -36,6 +37,10 @@ def get(hostname,
     :param use_rhceph: Whether or not to install RH Ceph on a RHEL machine or
                        the community distro.  Changes what host module is
                        returned for RHEL.
+    :params callbacks: A list of callables that accept one argument (the actual
+                       module that contains the connection) that will be
+                       called, in order at the end of the instantiation of the
+                       module.
     """
     conn = get_connection(
         hostname,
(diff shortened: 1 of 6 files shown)


trhoden:
The main thing that comes to mind with this is if there is a way we can keep from doing the check for if Ceph is installed over and over again...

I realize that sometimes we want it, sometimes we don't.

What if it becomes an optional argument to `hosts.get()`?  When we are initiating the connection with `hosts.get()`, we know whether we require to already be installed or not, so maybe adding it in by doing:

`hosts.get(..., ceph_required=True)` ?

Just thinking out loud.

ceph_deploy/hosts/__init__.py
```diff
@@ -70,6 +76,10 @@ def get(hostname,
     module.machine_type = machine_type
     module.init = module.choose_init(module)
     module.packager = module.get_packager(module)
+    # execute each callback if any
+    if callbacks:
+        for c in callbacks:
+            c(module)
```
trhoden: Awesome!

ceph_deploy/mon.py
```diff
@@ -257,7 +261,7 @@ def mon_create(args):
         try:
             # TODO add_bootstrap_peer_hint
             LOG.debug('detecting platform for host %s ...', name)
-            distro = hosts.get(host, username=args.username)
+            distro = hosts.get(host, username=args.username, callbacks=[packages.ceph_is_installed])
```
trhoden: This looks like the only `hosts.get()` call that didn't get split into multiple lines...  Not sure I care. :)

alfredodeza: I think it wasn't long enough. Not a problem if you prefer it split.

PR #346, opened 2015-08-14
mon.py:

blubanovic-penguin:
callbacks arg to get_connection() has no matching parameter, raises
exception at end of "ceph-deploy mon create-initial".

### DIFF
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -420,7 +420,6 @@ def mon_create_initial(args):
             host,
             username=args.username,
             logger=rlogger,
-            callbacks=[packages.ceph_is_installed]
         )
 
         while tries:


trhoden:
Hi Bill!

Nice catch -- guess you are running ceph-deploy master.  :)

A few things --  we would want a more detailed commit log description ("mon.py" doesn't tell me anything when looking at `git log`).

Second, the Ceph project requires a signed-off-by line when commiting code.  You can add this with `git commit -s --amend`.

But, to the problem at hand here -- I should have caught this.  @alfredodeza the `get_connection` piece seems to be a more direct path for a remote connect, skipping `hosts.get()` entirely.  I didn't even know that was there.  We *want* this check there for `mon_create_initial`, so removing the callback isn't the right fix.

We could either switch `get_connection` to `hosts.get()`, or we could see about adding the callback to `get_connection`.  Which would you prefer?  since the `ceph_is_installed` callback takes a module as input, it seems like we would have to switch over to `hosts.get()`, which is pretty trivial.

alfredodeza:
Ah, I apologize, I should've known better. Yes, this connection thing exists because it uses things that are non-default from `hosts.get`. We should implement the callback system there.
@blubanovic-penguin it should be just a couple of lines, do you mind adding them?

trhoden:
This is a terribly rude thing to do, but since `master` is currently broken, I need to get this fixed otherwise lots of CI tests for ceph are going to break over the weekend.

I would just merge patch from @blubanovic-penguin, and then implement fix on top of it, but can't without commit message fixup.  So I'm going to close to and fix separately.

Sorry.  :disappointed:

PR #348, opened 2015-08-28
install: add support for --dev-commit <sha1>

liewegas:
Adjusted centos, debian, fedora.

suse and rhel don't seem to pay attention to version_kind... :/

Signed-off-by: Sage Weil <sage@redhat.com>

### DIFF
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -63,7 +63,7 @@ def install(distro, version_kind, version, adjust_repos, **kw):
         key = 'autobuild'
 
     if adjust_repos:
-        if version_kind != 'dev':
+        if version_kind in ['stable', 'testing']:
             distro.packager.add_repo_gpg_key(gpg.url(key))
 
             if version_kind == 'stable':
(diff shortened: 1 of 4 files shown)


ceph_deploy/hosts/centos/install.py
```diff
@@ -98,6 +98,23 @@ def install(distro, version_kind, version, adjust_repos, **kw):
...
+                    machine=machine,
+                    version=version),
+                gpg.url(key),
+                adjust_repos=True,
+                extra_installs=False
+            )
+
+        else:
+            raise Exception('unrecognized version_kind %s' % version_kind)
+
```
trhoden: I hate to see all this duplicated code.  :)  Since it looks like the only difference is `ref` vs `sha1` in the URL, can we instead do something like:

```
if version_kind in ['dev', 'dev_commit']:
...

url = 'http://gitbuilder.ceph.com/ceph-rpm-centos{release}-{machine} basic/{type}/{version}/'.format(
    release=release.split(".", 1)[0],
    machine=machine,
    type='ref' if version_kind == 'dev' else 'sha1',
    version=version)

mirror_install(
    distro,
    url,
    gpg.url(key),
    adjust_repos=True,
    extra_installs=False
)

```

Same in Fedora.

trhoden:
Yay!  Thanks for making the change.  You improved on it even more. :)

PR #349, opened 2015-08-31
osd commands support "/dev/disk/by-path/".

osynge:
osd arguments can contain disk definitions by path.

/dev/disk/by-path/pci-0000:00:1f.2-ata-1.0

But these require : to delimit the host, disk and potentially journal.
Since argparse swallows \"' characters. But since \" and \' evaluate
to " and ' respectively we can use shlex to handle this correctly.

Hence a valid ceph-deploy command can now be:

```
ceph-deploy osd prepare --zap \
    ceph-node3:\'/dev/disk/by-path/pci-0000:00:1f.2-ata-1.0\'
```

or:

```
ceph-deploy osd prepare --zap \
    ceph-node3:\"/dev/disk/by-path/pci-0000:00:1f.2-ata-1.0\"
```

### DIFF
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -5,6 +5,7 @@ import os
 import re
 import sys
 import time
+import shlex
 from textwrap import dedent
 
 from cStringIO import StringIO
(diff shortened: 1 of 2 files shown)


osynge:
Am I correct in believing that the tests are failing for no good reason?

They seem to fail for me on both master and with this patch.

alfredodeza:
@osynge I would say they are failing correctly with assertion Errors. I would also suggest that this patch needs unit tests for the new logic/behavior added

alfredodeza:
Everything passes for me in master:

    $ tox
    ...
      py26: commands succeeded
      py27: commands succeeded
      flake8: commands succeeded
      congratulations :)
    $ git status
    On branch master
    Your branch is up-to-date with 'origin/master'.

osynge:
Running tox I get the output:

2029 passed, 16 skipped in 46.46 seconds 

But I do get some errors that I am not able to find python 2.6

alfredodeza:
@osynge you need to have Python 2.6 and 2.7 installed to be able to run all the environments that ceph-deploy is configured to execute with tox. Again: this PR has valid failures that you need to correct, and I would strongly suggest adding unit tests to support your changes.

osynge:
@alfredodeza I did have a bug, annoying issues with touple and list that had no impact, but made the test suite fail.

I followed your advice and added some tests to support the current behaviour and the changes.

Unrelated to this patch : would you recommend adding the same / similar tests for "ceph disk create" and "ceph disk activate" or is adding them for "ceph disk prepare" enough?

PR #350, opened 2015-09-02
install: add --test install option to install ceph-test package

liewegas:
We need this for the teuthology tests.

Don't include it in --all, since any normal person does not want it.

Signed-off-by: Sage Weil <sage@redhat.com>


alfredodeza:
This looks OK to me, but the tests will need to get updated. They are valid failures :)

alfredodeza:
This looks good to go to me. @trhoden ?

trhoden:
thanks for fixing the tests. :)

PR #353, opened 2015-09-10
jessie support

liewegas:


ceph_deploy/hosts/debian/__init__.py
```diff
@@ -18,7 +18,9 @@ def choose_init(module):
     """
     if distro.lower() == 'ubuntu':
         return 'upstart'
-    return 'sysvinit'
+    if not module.conn.remote_module.path_exists("/lib/systemd/system/ceph.target"):
+        return 'sysvinit'
+    return 'systemd'
```
trhoden: I know none of the LTS Ubuntu releases are on systemd yet, but since it is in 15.04, should we just switch this over to:

```
if ceph.target exists:
    systemd
elif ubuntu:
    upstart
else
    systemd
```

Thoughts?  That could be done later, of course.  What is here is fine.

ceph_deploy/hosts/debian/mon/create.py
```diff
@@ -1,42 +1,56 @@
...
 def create(distro, args, monitor_keyring):
-    logger = distro.conn.logger
     hostname = distro.conn.remote_module.shortname()
     common.mon_create(distro, args, monitor_keyring, hostname)
-    service = distro.conn.remote_module.which_service()
 
-    if not service:
-        logger.warning('could not find `service` executable')
-
-    if distro.init == 'upstart':  # Ubuntu uses upstart
```
trhoden: The sysvinit and systemd stuff look fine, but unless my eyes are playing tricks upstart got removed entirely.  So this would break on 14.04, for example.

liewegas: ah, so it did.  fixing..

ceph_deploy/hosts/debian/__init__.py
```diff
@@ -16,6 +16,8 @@ def choose_init(module):
 
     Returns the name of a init system (upstart, sysvinit ...).
     """
+    if module.conn.remote_module.path_exists("/lib/systemd/system/ceph.target"):
+        return 'systemd'
```
liewegas: actually, no, this isn't right either... since we install the systemd unit file regardless.  we need to detect what the host is running...  ceph-detect-init?

trhoden:
:+1: 

Thanks for the fixes/changes.

ceph_deploy/hosts/debian/__init__.py
trhoden: Since it is a runtime option, yes you are right.  It's something I've talked about with @dachary before that I know ceph-deploy needs to switch over to ceph-detect-init.

That is the longer-term fix, for sure.

For now, the only time we would get in trouble is people running Infernalis RC on Debian and changing the init at runtime?  Or is that even on older distros (14.04) that Infernalis packages are going to install `ceph.target`?

liewegas: yeah, the problematic cases are jessie with non-systemd and ubuntu 
15.04...

PR #355, opened 2015-09-14
Refactor handling of init systems

codenrhoden:
The idea here is to create and use classes with a common interface to abstract working with the init systems.  Should simplify the code immensely.

This will be a pre-req to fixing http://tracker.ceph.com/issues/13087, which has us use the new `ceph-detect-init` tool when it is present.


trhoden:
I think this in a good state to review now.

I've tested on CentOS 7, using both Hammer release, and latest Infernalis dev packages.  This covered both sysvinit and systemd.

I did the same thing on Trusty, and that successfully gets upstart on each.

Using the latest Infernalis packages, I did get an error when trying activate file-based OSDs on both Trusty and CentOS 7::

```
[trusty1][WARNIN] 2015-09-15 23:30:57.355914 7f60eede0980 -1 filestore(/tmp/osd1) mkfs: write_version_stamp() failed: (13) Permission denied
[trusty1][WARNIN] 2015-09-15 23:30:57.355938 7f60eede0980 -1 OSD::mkfs: ObjectStore::mkfs failed with error -13
[trusty1][WARNIN] 2015-09-15 23:30:57.357761 7f60eede0980 -1  ** ERROR: error creating empty object store in /tmp/osd1: (13) Permission denied
```

This seems more to do with running the daemons as the ceph users than anything to do with the init system, though.

I should still do a couple tests using RGW and MDS, but it's looking good so far.

@alfredodeza @liewegas feel free to take a look.  There is additional clean up that could be done, but this is all the obvious stuff.  It is nice to not have distro-specific versions of the "mon create" stuff anymore. :)

Once this is in, I can work on switching over to `ceph-detect-init` when its presence is detected.

liewegas:
On Tue, 15 Sep 2015, Travis Rhoden wrote:
> [trusty1][WARNIN] 2015-09-15 23:30:57.355914 7f60eede0980 -1 filestore(/tmp/osd1) mkfs: write_version_stamp() failed: (13) Permission denied
> [trusty1][WARNIN] 2015-09-15 23:30:57.355938 7f60eede0980 -1 OSD::mkfs: ObjectStore::mkfs failed with error -13
> [trusty1][WARNIN] 2015-09-15 23:30:57.357761 7f60eede0980 -1  ** ERROR: error creating empty object store in /tmp/osd1: (13) Permission denied

if it's dcrypt, that's http://tracker.ceph.com/issues/13000

> This seems more to do with running the daemons as the ceph users than anything to do with the init system, though.

yeah

trhoden:
FWIW, the permissions error is because I was using directories for OSDs (which I rarely do) and created them as root.  

If I did...

```
chown -R ceph:ceph /tmp/osd1
ceph-deploy osd prepare <host>:/tmp/osd1
ceph-deploy osd activate <host>:/tmp/osd1
```

things worked just fine.  So as part of Infernalis, we'll probably need to update docs to say that if you are using directories for OSD testing, you need to set the right perms.  :)  Block devices are all taken care of automatically by the tooling.  I wonder if ceph-disk should do something similar when pointed at a directory?

trhoden:
RGW and MDS services seem correct on Trusty/Upstart.  

I had to push one more commit to fix RGW on sysvinit.  I wasn't properly handling that the 'ceph' and 'ceph-radosgw' init scripts have completely different options (like `-c`, or that the rgw script can't take a daemon name).

That's all fixed now.

trhoden:
@dmick @alfredodeza care to take a look at this one?

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+    name = 'sysvinit'
+
+    def __init__(self, remote_conn):
+        super(SysV, self).__init__(remote_conn)
+        self.service_exe = self.remote_conn.remote_module.which_service()
+
+    def _get_init_script(self, service):
+        return 'ceph-radosgw' if service == 'ceph-radosgw' else 'ceph'
+
+    def _build_service_command(
```
alfredodeza: I didn't see any tests for this one. It seems like it is one of the few that has a lot of behavior (maybe there are tests for it already?)

It might get tricky to test it out because the `InitSystem` class has eager behavior in `__init__` (it connects remotely to find out what the hostname is).

If that results in making it difficult to test maybe `hostname` should be a `@property` so that is only called/used when needed

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def enable(self, service, **kw):
+        """Enable service on remote node"""
+        raise NotImplementedError()
+
+    def disable(self, service, **kw):
+        """Disable service on remote node"""
+        raise NotImplementedError()
+
+    def status(self, service, **kw):
```
osynge: I feel that status method here maybe redundant. I wonder if it should be replaced with some code to establish if the service is running which returns a boolean true or false.

Admittedly in master so far we do not have a use case yet for status, but I am using it in the SUSE rgw code, I use similar code to this to establishing if apache is running.

For a tested implementation for systemd:

https://github.com/osynge/ceph-deploy/commit/09dc6213e645ee2fe7bbc16d82b21cc350f2ceb3

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+# Right Thing.
+SYSV_SERVICES = {
+    'ceph-mon': 'mon',
+    'ceph-mds': 'mds'
+}
+
+UPSTART_SERVICES = {
+    'ceph-radosgw': 'radosgw'
+}
+
```
osynge: I think that SYSV_SERVICES and  UPSTART_SERVICES mappings are in the wrong place here.

These are distribution specific mappings, rather than init system specific.

This would change the interface a little but will make the code more neutral to SUSE, REDHAT, debian differences etc.

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+
+    name = 'sysvinit'
+
+    def __init__(self, remote_conn):
+        super(SysV, self).__init__(remote_conn)
+        self.service_exe = self.remote_conn.remote_module.which_service()
+
+    def _get_init_script(self, service):
+        return 'ceph-radosgw' if service == 'ceph-radosgw' else 'ceph'
```
osynge: This looks like a syntax error.

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+                'chkconfig',
+                init_script,
+                'off'
+            ]
+        )
+
+    def status(self, service, **kw):
+        name = kw.pop('name', self.hostname)
+        cmd = self._build_service_command(service, 'status', name)
+        self._run(cmd)
```
osynge: No return value, so not sure what is the need for this function?

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+                'initctl',
+                'status',
+                '{service}'.format(
+                    service=UPSTART_SERVICES.get(service, service)
+                ),
+                'cluster={cluster}'.format(cluster=cluster),
+                'id={name}'.format(name=name)
+            ]
+        )
+
```
osynge: Again no return value for this method

ceph_deploy/util/init_systems.py
```diff
@@ -0,0 +1,246 @@
...
+    def status(self, service, **kw):
+        name = kw.pop('name', self.hostname)
+        self._run(
+            [
+                'systemctl',
+                'status',
+                '{service}@{name}'.format(service=service, name=name),
+            ]
+        )
```
osynge: Again no return value

ceph_deploy/hosts/fedora/__init__.py
```diff
@@ -19,9 +18,9 @@ def choose_init(module):
     """
 
     if not module.conn.remote_module.path_exists("/usr/lib/systemd/system/ceph.target"):
-        return 'sysvinit'
+        return init_systems.SysV(module)
```
osynge: Am I correct in thinking this is now returning a class?

This would also be a very good time to "inject" the service mappings, for fedora rather than have them hard coded in "init_systems"

If so the comments for all OS need changing.

ceph_deploy/util/init_systems.py
osynge: I am thinking along the lines of https://github.com/ceph/ceph-deploy/pull/318

ceph_deploy/hosts/fedora/__init__.py
osynge: I am thinking along the lines of https://github.com/ceph/ceph-deploy/pull/318

trhoden:
Okay, so here is what I have done...

@alfredodeza I added unit tests to cover the eventual command invocations.  Right now it apepars I have used some calls in `Mock` that are not available in py2.6, so I've got to look into that (that's what the test failure is -- all is fine in py2.7).

To address a few of @osynge comments, I have removed the `stop`, `status`, and `disable` methods for the init classes.  These are not used anywhere in the code right now, so adding them is just bloat and leads to confusion.  And as Owen pointed out, some of them weren't even usable.

I need to address the test errors, and also Owen's comments about service mappings and PR #318 still...

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist = py26, py27, flake8
 
 [testenv]
 deps=
-  pytest
+  pytest<2.8.0
```
alfredodeza: good catch

trhoden: Yeah, it was the import of `call` from Mock.  See https://github.com/pytest-dev/pytest/issues/1044 (which is a dup of another one).

ceph_deploy/util/init_systems.py
trhoden: I don't think any of these need return values.  the `_run()` method is calling `remoto.process.run()`, which will raise an exception (by default) if the the command returns non-zero.

If any of our commands on remote nodes return non-zero, we have ceph-deploy halt it's execution.  I'm not looking to change behavior here.

That being said, this does highlight that I call `return` in `InitSystem._run()`, which doesn't make any sense.  I can remove that.

trhoden: It's been removed.

ceph_deploy/util/init_systems.py
trhoden: @alfredodeza Thanks for pointing out the eager behavior in `__init__`.  I've removed that in the two places I had it and used `@property` instead.  A definite improvement.

ceph_deploy/util/init_systems.py
osynge: My comment is no longer relevant as it was for the "status" command which you have since removed.

ceph_deploy/util/init_systems.py
trhoden: @osynge You are absolutely right.  There are a couple of things.  First, what I've done is slightly confusing because in the sysvinit case I've conflated a service with the ceph init scripts capability to take an optional daemon type/instance.

For example, there are really two ceph _services_, `ceph` and `ceph-radosgw`.  However, the `ceph` service (init script) can also take a daemon to act on (like `mds` `osd` `mon`).  So both of these are valid:

`service ceph start`
`service ceph start mon.hostname`

The goal with `SYSV_SERVICES` here is that when the function sees a request to "start ceph-mon", it knows to create a command along the lines of `service ceph start mon.<x>`.  This behavior is init system specific, not distro specific.  The `SYSV_SERVICES` dictionary name does not make this behavior clear at all.

On top of that, the Debian-derived _distros_ (*.deb* packages) use a `radosgw` service instead of `ceph-radosgw`, and we correctly capture that Ubuntu (Upstart) but not on Debian today (http://tracker.ceph.com/issues/12170).  The work I've done here would perpetuate that error.

As it turns out, today is my last day with Red Hat.  I hope to see this PR through, but I'm not sure if I can fully address the issue in http://tracker.ceph.com/issues/12170 and #318.  I need to rebase this to make it merge clean, but you are right that something else needs to happen.

I *think* it might be as simple as calling `self.remote_info.service_mapping()` in the places necessary in the needed places.  That would require an interface change.

What I'm getting at is that I"m not sure if I can get the change into this PR or not.  But I wanted to detail would be necessary to fix it.  I don't think this change takes any steps backwards, but it doesn't fix the known issue that you correctly point out.

codenrhoden:
Oh man - I never realized this never went in. That's a shame, it was a big improvement! :)

I suppose this should just be closed at this point.

liewegas:
can it be rebased?

osynge:
@codenrhoden I think this could be rebased and made to work again and is a very big improvement.

codenrhoden:
@liewegas @osynge I'm sure it can be. I just need to go through any fixes/improvements from the last year to see if there is anything that needs to be added.

I'll try to get to that over the next week or two.

I'm going to be working on some Ceph related stuff over the next few weeks, so I'm happy to help out here!

vasukulkarni:
@codenrhoden sorry this seems to have become old,  the newer centos and ubuntu is only systemd now and the current logic although simple handles it better and is well tested.

codenrhoden:
well whether it makes sense or not, I'll take a stab at it. I've just pushed an initial rebase, but no testing has been done at this point.

vasukulkarni:
running against *this* branch on hammer, jewel 

http://pulpito.ceph.com/vasu-2016-10-03_16:53:30-ceph-deploy-hammer---basic-vps/
http://pulpito.ceph.com/vasu-2016-10-03_16:54:03-ceph-deploy-jewel---basic-vps/

codenrhoden:
@vasukulkarni: thanks for the links. I haven't done much testing yet, so I don't expect things to work. However, the errors in Pulpito are not code related. They are all failing when doing `git clone -b RM-13087 git://git.ceph.com/ceph-deploy.git`. The problem is there is no `RM-13087` branch at git.ceph.com/ceph-deploy.git.  That branch only exists in my (`codenrhoden`) fork.

Not sure how Pulpito normally handles that.

If the branch can be overriden, I believe `git clone -b refs/pull/355/head ...` might work, but only if refs/pull/* are also copied over to git.ceph.com.

codenrhoden:
We'll see what Jenkins thinks of the latest changes. I was able to get all tests passing on py27 and py34 plus flake8 with latest changes (updating pytest and mock version requirements). I wasn't easily able to get a py26 env installed on CentOS 7.2. I guess it's no longer in EPEL?

codenrhoden:
```
  py26: commands succeeded
  py27: commands succeeded
ERROR:   py34: InterpreterNotFound: python3.4
  flake8: commands succeeded
```

So, if Python 3.4 is going to be tested, you should probably add it to the Jenkins nodes. :)

vasukulkarni:
@codenrhoden that failure as you said is because its in your git repo which i didn't notice , it only looks for branh in ceph repo https://github.com/ceph/ceph-qa-suite/blob/master/tasks/ceph_deploy.py#L66, If you can create the branch in ceph-deploy repo it will be easy to get some runs on hammer/jewel

alfredodeza:
ceph is systemd only now

PR #357, opened 2015-09-18
download.ceph.com has https now

liewegas:


dmick:
Did you add the SSL config  this morning?

liewegas:
yup!  the ssl cert delay was me.. i had to clink teh link in teh email dh 
sent

ceph_deploy/hosts/debian/install.py
```diff
@@ -30,11 +30,11 @@ def install(distro, version_kind, version, adjust_repos, **kw):
         distro.packager.add_repo_gpg_key(gpg.url(key, protocol=protocol))
 
         if version_kind == 'stable':
-            url = 'http://download.ceph.com/debian-{version}/'.format(
+            url = 'https://download.ceph.com/debian-{version}/'.format(
```
trhoden: I might need to see if the SSL cert in use is valid "out of the box" on wheezy.  Right above this code is a special case where we fallback to http on wheeze to access to the GPG key, because otherwise we get invalid cert errors.

trhoden: yeah, unfortunately https doesn't work on wheezy:

```
[vagrant-debian-wheezy][DEBUG ] add deb repo to /etc/apt/sources.list.d/
[vagrant-debian-wheezy][INFO  ] Running command: env DEBIAN_FRONTEND=noninteractive DEBIAN_PRIORITY=critical apt-get --assume-yes -q update
[vagrant-debian-wheezy][WARNIN] E: The method driver /usr/lib/apt/methods/https could not be found.
[vagrant-debian-wheezy][ERROR ] RuntimeError: command returned non-zero exit status: 100
[ceph_deploy][ERROR ] RuntimeError: Failed to execute command: env DEBIAN_FRONTEND=noninteractive DEBIAN_PRIORITY=critical apt-get --assume-yes -q update
```

Looks like it's not supported at all via apt.  Weird!

I tried it with http and it worked.

dmick: apt requires a plugin in a separate package:  apt-transport-https on Trusty.  Try and see if that package is installed.

dmick: ...and I guess we should do something to make sure that package is installed in general in the tool.

trhoden: `apt-transport-https` is present on wheezy.  Let's just force install it right where we install `ca-certificates`?  Sounds like the way to go.

dmick: :thumbsup:

trhoden: @liewegas Just to follow up here.  I think we are looking for one change, which is to change:

`distro.packager.install('ca-certificates')`

into:

`distro.packager.install(['ca-certificates', 'apt-transport-https'])`

liewegas: Added!

trhoden:
Great, that seems to do what is intended -- make apt able to use https.

but it still looks like wheezy does not like our SSL cert. :(

```
[vagrant-debian-wheezy][DEBUG ] Err https://download.ceph.com wheezy/main amd64 Packages
[vagrant-debian-wheezy][DEBUG ]   server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
[vagrant-debian-wheezy][WARNIN] W: There is no public key available for the following key IDs:
[vagrant-debian-wheezy][WARNIN] 9D6D8F6BC857C906
[vagrant-debian-wheezy][WARNIN] W: There is no public key available for the following key IDs:
[vagrant-debian-wheezy][WARNIN] 7638D0442B90D010
[vagrant-debian-wheezy][WARNIN] W: Failed to fetch https://download.ceph.com/debian-hammer/dists/wheezy/main/binary-amd64/Packages  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
[vagrant-debian-wheezy][WARNIN]
[vagrant-debian-wheezy][WARNIN] E: Some index files failed to download. They have been ignored, or old ones used instead.
```

**server certificate verification failed**

I think that's why we fall back to http on wheezy for installing the GPG key (via `wget`).

trhoden:
It works!

dmick:
I think maybe the change to gitbuilder.ceph.com was unintentional?  It doesn't do, and hasn't ever done, https

liewegas:
https://github.com/ceph/ceph-deploy/pull/364

PR #359, opened 2015-09-24
install: functions "uninstall" and "purge" merged to function "remove"

osynge:
Before this patch functions "uninstall" and "purge" are almost identical.
Since the only difference is the logging messages, and when the
distro.uninstall() function is called a boolean parameter set both these
functions are merged into a single command "remove" with the parameter
purge.

### DIFF
--- a/ceph_deploy/install.py
+++ b/ceph_deploy/install.py
@@ -293,12 +293,15 @@ def install_repo(args):
 
         custom_repo(distro, args, cd_conf, rlogger, install_ceph=False)
 
-
-def uninstall(args):
+def remove(args, purge):
     LOG.info('note that some dependencies *will not* be removed because they can cause issues with qemu-kvm')
     LOG.info('like: librbd1 and librados2')
+    remove_action = 'Uninstalling'
+    if purge:
+        remove_action = 'Purging'
     LOG.debug(
-        'Uninstalling on cluster %s hosts %s',
+        '%s on cluster %s hosts %s',
+        remove_action,
         args.cluster,
         ' '.join(args.host),
         )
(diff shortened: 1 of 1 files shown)


osynge:
rebased to validate CI tests.

osynge:
Fixed small bug in logging.

ceph_deploy/install.py
```diff
@@ -293,12 +293,15 @@ def install_repo(args):
 
         custom_repo(distro, args, cd_conf, rlogger, install_ceph=False)
 
-
-def uninstall(args):
-    LOG.info('note that some dependencies *will not* be removed because they can cause issues with qemu-kvm')
+def remove(args, purge):
+    LOG.info('note that some dependencies *will not* be d because they can cause issues with qemu-kvm')
```
alfredodeza: seems like a typo here, why did this line got changed anyway?

osynge: Ops no idea, will fix now.

osynge:
Thanks for spotting the typo.

ceph_deploy/install.py
```diff
@@ -312,35 +315,15 @@ def uninstall(args):
             use_rhceph=True)
         LOG.info('Distro info: %s %s %s', distro.name, distro.release, distro.codename)
         rlogger = logging.getLogger(hostname)
-        rlogger.info('uninstalling Ceph on %s' % hostname)
-        distro.uninstall(distro)
+        rlogger.info('%s Ceph on %s' % (remove_action, hostname))
+        distro.uninstall(distro, purge)
```
trhoden: Would you be open to changing this line to: `distro.uninstall(distro, purge=purge)`?  `purge` is a kwarg here, and I would prefer to see that called out explicitly rather than it looking like a positional argument.

trhoden:
This looks great.  One small request, otherwise LGTM.

ceph_deploy/install.py
osynge: good point we need more clarity here I have just made the change requested.

osynge:
Hope the change addresses things clearly.

PR #362, opened 2015-09-26
install --no-check-packages-signatures for test purposes

dachary:
http://tracker.ceph.com/issues/13249

When installing from repositories that are created dynamically within a
lab, verifying the packages signatures can be skipped. The
--no-check-packages-signatures configures the repositories to not
require package signatures verification during installation.

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -65,17 +65,25 @@ def set_apt_priority(fqdn, path='/etc/apt/preferences.d/ceph.pref'):
 
 
 def set_repo_priority(sections, path='/etc/yum.repos.d/ceph.repo', priority='1'):
+    set_repo_key_value(sections, path, 'priority', priority)
+
+
+def set_repo_gpgcheck(sections, path='/etc/yum.repos.d/ceph.repo', gpgcheck='1'):
+    set_repo_key_value(sections, path, 'gpgcheck', gpgcheck)
+
+
+def set_repo_key_value(sections, path, key, value):
     Config = ConfigParser.ConfigParser()
     Config.read(path)
     Config.sections()
     for section in sections:
         try:
-            Config.set(section, 'priority', priority)
+            Config.set(section, key, value)
         except ConfigParser.NoSectionError:
             # Emperor versions of Ceph used all lowercase sections
             # so lets just try again for the section that failed, maybe
             # we are able to find it if it is lower
-            Config.set(section.lower(), 'priority', priority)
+            Config.set(section.lower(), key, value)
 
     with open(path, 'wb') as fout:
         Config.write(fout)
--- a/ceph_deploy/install.py
+++ b/ceph_deploy/install.py
@@ -183,6 +183,7 @@ def install(args):
                 version,
                 args.adjust_repos,
                 components=components,
+                no_check_packages_signatures=args.no_check_packages_signatures,
             )
 
         # Check the ceph version we just installed
@@ -597,6 +598,12 @@ def make(parser):
         dest='gpg_url',
         help='specify a GPG key URL to be used with custom repos\
                 (defaults to ceph.com)'
+    )
+
+    parser.add_argument(
+        '--no-check-packages-signatures',
+        action='store_true',
+        help='do not check the package signatures',
     )
 
     parser.set_defaults(
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -105,6 +105,10 @@ def install(distro, version_kind, version, adjust_repos, **kw):
         # set the right priority
         logger.warning('ensuring that /etc/yum.repos.d/ceph.repo contains a high priority')
         distro.conn.remote_module.set_repo_priority(['Ceph', 'Ceph-noarch', 'ceph-source'])
+        if kw['no_check_packages_signatures']:
+            distro.conn.remote_module.set_repo_gpgcheck(['Ceph', 'Ceph-noarch', 'ceph-source'],
+                                                        '/etc/yum.repos.d/ceph.repo',
+                                                        '0')
         logger.warning('altered ceph.repo priorities to contain: priority=1')
 
     if packages:
--- a/ceph_deploy/hosts/debian/install.py
+++ b/ceph_deploy/hosts/debian/install.py
@@ -27,7 +27,8 @@ def install(distro, version_kind, version, adjust_repos, **kw):
         protocol = 'https'
         if codename == 'wheezy':
             protocol = 'http'
-        distro.packager.add_repo_gpg_key(gpg.url(key, protocol=protocol))
+        if not kw['no_check_packages_signatures']:
+            distro.packager.add_repo_gpg_key(gpg.url(key, protocol=protocol))
 
         if version_kind == 'stable':
             url = 'http://download.ceph.com/debian-{version}/'.format(
@@ -54,9 +55,12 @@ def install(distro, version_kind, version, adjust_repos, **kw):
 
     # TODO this does not downgrade -- should it?
     if packages:
+        extra_install_flags=['-o', 'Dpkg::Options::=--force-confnew']
+        if kw['no_check_packages_signatures']:
+            extra_install_flags.append('--force-yes')
         distro.packager.install(
             packages,
-            extra_install_flags=['-o', 'Dpkg::Options::=--force-confnew']
+            extra_install_flags=extra_install_flags
         )
 
 


smithfarm:
Just an English grammar nitpick: it should be `--no-check-package-signatures`.

dachary:
@smithfarm fixed, thank you !

ceph_deploy/hosts/remotes.py
```diff
@@ -65,17 +65,25 @@ def set_apt_priority(fqdn, path='/etc/apt/preferences.d/ceph.pref'):
...
+    set_repo_key_value(sections, path, 'gpgcheck', gpgcheck)
+
+
+def set_repo_key_value(sections, path, key, value):
     Config = ConfigParser.ConfigParser()
     Config.read(path)
     Config.sections()
     for section in sections:
         try:
-            Config.set(section, 'priority', priority)
```
alfredodeza: wow, this is a good catch here

alfredodeza:
I don't think we should support something like this with a CLI flag. ceph-deploy has extensive support for repositories and custom configurations there. To avoid a GPG check on a YUM repository for example, you would need to add the following to the repository section in `cephdeploy.conf`:

    gpgcheck=0

alfredodeza:
I also didn't see anything for Debian in this PR. And there is currently no way to specify skipping the GPG check for them with either flags or configuration, so we should probably check for `gpgcheck=0` for debian repos as well (vs. a flag)

dachary:
The debian part was added. Note that this is used in the gitbuilder code path which does not use the templates embeded in ceph-deploy.conf. Teuthology currently assumes all configuration to ceph-deploy can be provided via the command line, reason why I went this way. Using the configuration file instead would require more work on ceph-deploy (so that the function can use a configuration file instead of what is done currently) and more work on teuthology (to create the configuration file and upload it in addition to calling the ceph-deploy cli to achieve the desired combination of options).

trhoden:
I don't think I'm opposed to this in general, however it does solidify a departure from what our initial/main priorities are with ceph-deploy.  This PR and #363 together really make ceph-deploy a tool for dev testing, not just a tool for installing POC clusters from upstream packages.

I realize this is a completely blurry and arbitrary distinction, as we already use ceph-deploy with Teuthology, and have been saying that we want to do that even more (and this helps with that).  I also already merged #348, which is in the same category.  On those grounds, it only makes sense to take this in, I think.

I know that Teuthology does not make use of ceph-deploy's .conf file at all.

We had always supported only installing packages from on of *our* repos -- be it ceph.com, gitbuilder, or a downstream distros.  These have always been signed, so that's why no support.  What is the use case for installing unsigned packages?  Is it pure developers?

ceph_deploy/install.py
```diff
@@ -599,6 +600,12 @@ def make(parser):
                 (defaults to ceph.com)'
     )
 
+    repo.add_argument(
+        '--no-check-packages-signatures',
+        action='store_true',
+        help='do not check the package signatures',
+    )
+
```
trhoden: I don't think adding this arg to `repo` is what you want.  `repo` is a mutex group between `--adjust-repos | --no-adjust-repos | --repo`.  This would add the flag into that group.  Since this is seemingly relevant to all flags under `ceph-deploy install ...`, I think it would be `parser.add_argument()` instead.

trhoden:
I also wonder if we should do an equivalent option in https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/repo.py, if we are really going to flesh this feature out fully.

dachary:
> What is the use case for installing unsigned packages? Is it pure developers? 

The use case I'm addressing is packages that are bulit on the fly when running a teuthology suite. For instance, when a developer runs the following from the command line: 
<pre>
teuthology-openstack --os-type centos --os-version 7.0 --ceph wip-mybranch --ceph-git-url http://github.com/my/ceph --suite mysuite
</pre>
It:

* creates a virtual machine
* creates a teuthology cluster in the virtual machine
* adds the buildpackages task to mysuite
* when the buildpackages task runs (before the install task or before the ceph-deploy task) it
** creates an http server if it does not already exists and makes it the gitbuilder host
** makes the packages for centos-7.0 and uploads them to the http server, if they do not already exists for the branch wip-mybranch
* runs mysuite with this transparent gitbuilder replacement

dachary:
> I don't think adding this arg to repo is what you want.

Oops, my bad. Fixing.

dachary:
@trhoden I'm willing to redo this implementation from scratch to better fit the long term goal, just let me know in which direction to go. I think it will be inconvenient for teuthology to run ceph-deploy by creating a configuration file in addition to the arguments / options on the command line. Other than this, no concern :-)

dachary:
@trhoden my understanding of the teuthology ceph-deploy task is that it is designed to verify that ceph-deploy can successfully create a cluster with the packages created from a given branch. In that sense ceph-deploy is tested by teuthology more than it uses ceph-deploy for tests.

dachary:
@trhoden I'll have time this week to refactor this change if you think it necessary.

liewegas:
I think this looks ok in principle; I have no opinion on implementation.  I suspect @trhoden's comment about the command line option needs fixing though?

dachary:
> trhoden's comment about the command line option needs fixing though?

It was fixed, not sure why the github interface suggests otherwise.

trhoden:
@dachary 

I think it definitely comes down to whether we want to add a bunch more flags to `ceph-deploy install`, or use the `ceph-deploy repo` command instead.

As it turns out, I think the `ceph-deploy repo` command can already handle the needed scenario in this PR and also in #363.  Here's the help output:

```
$ ceph-deploy repo --help
usage: ceph-deploy repo [-h] [--repo-url REPO_URL] [--gpg-url GPG_URL]
                        [--remove]
                        REPO-NAME HOST [HOST ...]

Repo definition management

positional arguments:
  REPO-NAME            Name of repo to manage. Can match an entry in
                       cephdeploy.conf
  HOST                 host(s) to install on

optional arguments:
  -h, --help           show this help message and exit
  --repo-url REPO_URL  a repo URL that mirrors/contains Ceph packages
  --gpg-url GPG_URL    a GPG key URL to be used with custom repos
  --remove, --delete   remove repo definition on remote host
```

If you use the command to add a repo definition with the `-repo-url` command, the `--gpg-url` flag is optional.  If you *don't* pass it, GPG checking is automatically disabled.  This happens here: https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/util/pkg_managers.py#L109-L113


So what I would propose is that instead of doing

```
ceph-deploy install --gitbuilder-host {url} --dev-commit {sha1} --no-check-packages <node>
```

I would instead do:

```
ceph-deploy repo --repo-url {url} <node>
ceph-deploy install --dev-commit {sha1} <node>
```

The caveats here are that Teuthology can be modified to use the `ceph-deploy repo` command, and that there does not exist on the nodes a conflicting repo definition.

I think it makes sense in the long term to have any needed repo modification capabilities in ceph-deploy be done through a `repo` command, but that is just my opinion.  I understand others may feel differently since it introduces a new command.  Internal to ceph-deploy, I would also want to see the repos that are created in the _default_ `install` scenarios re-use the `distro.packager.add_repo` logic rather than the current stuff that happens in https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/hosts/centos/install.py#L124-L131, for example.

dachary:
@trhoden I took a quick look at the code involved in the ```repo``` sub-command. Unless I'm mistaken it is different from the code path used for gitbuilders. Am I missing something ?

trhoden:
@dachary No I don't think you've missed anything.  I forgot how involved the gitbuilder URL structure is, and that you probably won't want to provide the full format of `'http://gitbuilder.ceph.com/ceph-rpm-centos{release}-{machine}-basic/{sub}/{version}/'` to something like `--repo-url`.

I did intend to point out the use of `--no-adjust-repos` in my previous example, and while it would be required, it would be get you all the way there unless you actually passed the full URL as in this comment.

trhoden:
so, I think that it would come down to enhancing the `ceph-deploy repo` command to take a `--dev` flag into indicate a gitbuilder URL structure, so that it could tack on the appropriate path.

That's a tough call on which path to take.

dachary:
@trhoden if I'm not mistaken both path require non trivial refactoring. Or is there a simple solution I'm not seeing ?

dachary:
I think I understand how this should be done now, thanks for explaining.

trhoden:
@dachary I think you are right that it it is not so simple/trivial.  I am not sure what the best course is, cost/benefit wise.  And, my opinion matters less now.  :)  I think a lot of ceph-deploy maintenance is going to fall back to @alfredodeza.

PR #363, opened 2015-09-26
install --gitbuilder-host alternative to gitbuilder.ceph.com

dachary:
To be used with --gitbuilder-host gitbuilder.myself.net
--dev-commit 6f11b82f3007819b58f143ad4b4911c4fb766ab9. It is primarily
useful in the context of the ceph_deploy.py task, when a full gitbuilder
mirror is available, with the exact same directory layout.

http://tracker.ceph.com/issues/13254 Fixes: #13254

### DIFF
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -89,11 +89,13 @@ def install(distro, version_kind, version, adjust_repos, **kw):
             logger.info('repo file will be created manually')
             mirror_install(
                 distro,
-                'http://gitbuilder.ceph.com/ceph-rpm-centos{release}-{machine}-basic/{sub}/{version}/'.format(
+                'http://{gitbuilder_host}/ceph-rpm-centos{release}-{machine}-basic/{sub}/{version}/'.format(
                     release=release.split(".", 1)[0],
                     machine=machine,
                     sub='ref' if version_kind == 'dev' else 'sha1',
-                    version=version),
+                    version=version,
+                    gitbuilder_host=kw['gitbuilder_host'],
+                ),
                 gpg.url(key),
                 adjust_repos=True,
                 extra_installs=False
--- a/ceph_deploy/hosts/debian/install.py
+++ b/ceph_deploy/hosts/debian/install.py
@@ -36,11 +36,12 @@ def install(distro, version_kind, version, adjust_repos, **kw):
         elif version_kind == 'testing':
             url = 'http://download.ceph.com/debian-testing/'
         elif version_kind in ['dev', 'dev_commit']:
-            url = 'http://gitbuilder.ceph.com/ceph-deb-{codename}-{machine}-basic/{sub}/{version}'.format(
+            url = 'http://{gitbuilder_host}/ceph-deb-{codename}-{machine}-basic/{sub}/{version}'.format(
                 codename=codename,
                 machine=machine,
                 sub='ref' if version_kind == 'dev' else 'sha1',
                 version=version,
+                gitbuilder_host=kw['gitbuilder_host'],
                 )
         else:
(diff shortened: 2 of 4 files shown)


dachary:
@trhoden would you mind taking a look at this pull request ? Please unassign yourself if not. Not quite sure about the protocol as this is my first contribution to ceph-deploy :-)

alfredodeza:
In this case, I still think that there is no need for adding another flag. The same can be accomplished using the `cephdeploy.conf` file with whatever URL the task wants to use.

Have you read the docs on ceph-deploy configuration files and not found them appropriate here?

https://github.com/ceph/ceph-deploy/blob/master/docs/source/conf.rst#ceph-deploy-configuration

The reason the configuration file was created was to give support to all these variants without the need to add more flags. For an automated system this should be easy to implement as the `cephdeploy.conf` file follows the INI format.

dachary:
@alfredodeza https://github.com/ceph/ceph-deploy/blob/master/docs/source/conf.rst#ceph-deploy-configuration is not used by the code path that implements the gitbuilder naming conventions.

alfredodeza:
@dachary right, but you could write the actual gitbuilder URL to a cephdeploy.conf file and not depend on the code that does that

dachary:
@alfredodeza I could also install the packages without ceph-deploy, but both options would be more redundant code / effort than substituting a hardcoded value.

dachary:
I don't see a substantial difference between this pull request and https://github.com/ceph/ceph-deploy/pull/348/files

trhoden:
Yeah I probably opened the door for all this with #348.  I considered that PR to be just an easy shortcut to referencing a different set of packages in a place that we already could point to.

My longer term goal had been to have most of the "repo" handling moved into ceph_deploy/repo.py, and the corresponding `ceph-deploy repo` commands.  Those can already take custom repo URLs, and could be enhanced with flags for GPG (a la #362) or whatever else we wanted to toggle for repos via CLI.

I'm hesitant to expose the notion/concept of a "gitbuilder" to the user.  The gitbuilders do have a unique URL structure to them, for sure, and installing with `--dev` triggers logic to know about them, but otherwise a gitbuilder is just a URL to hit.

Are there going to be multiple instances of gitbuilders now?

dachary:
@trhoden thanks for explaining the long term goal. 

In the context of the OpenStack backend for teuthology, there are indeed multiple repositories structured to look exactly like the gitbuilder. There no longer is a unique gitbuilder repository. They are not created by gitbuilder but otherwise look the same. The teuthology tasks rely on the gitbuilder naming convention extensively, it's not just about ceph-deploy.

alfredodeza:
@trhoden @dachary I wouldn't rely on a URL structure for *gitbuilders*, there is a lot of work already taking place to be able to get away from them and implement repositories that match a different URL structure altogether.

This alone should be a heftier reason on why work towards relying on a URL structure like the ones in gitbuilders is not a good long-term solution and why I think just using cephdeploy.conf to handle them in the mean time is the correct approach here.

alfredodeza:
@dachary if you are already creating a repository from scratch why do you need that to match a gitbuilder structure? What is the reason you can't use a completely different url that ceph-deploy can consume?

ceph-deploy already can handle arbitrary urls and these can be provided on the CLI even, mind explaining a bit the logic behind trying to match the gitbuilder structure?

dachary:
> why do you need that to match a gitbuilder structure? 

Because the teuthology tasks rely on the gitbuilder naming convention extensively, it's not just about ceph-deploy.

dachary:
@trhoden ping ?

liewegas:
:+1: This looks right to me.  This lets us mirror gitbuilder.ceph.com (or use a proxy) for other test environments.

PR #366, opened 2015-10-19
apt: no gpg key is installed allows unauthenticated packages

dachary:
Signed-off-by: Loic Dachary <ldachary@redhat.com>

### DIFF
--- a/ceph_deploy/util/pkg_managers.py
+++ b/ceph_deploy/util/pkg_managers.py
@@ -235,10 +235,21 @@ class Apt(PackageManager):
         cmd = ['apt-key', 'add', gpg_file]
         self._run(cmd)
 
+    def disable_signatures_check(self):
+        self.remote_conn.logger.info(
+            "Append APT::Get::AllowUnauthenticated 1 to /etc/apt/apt.conf"
+        )
+        self.remote_conn.remote_module.append_to_file(
+            '/etc/apt/apt.conf',
+            'APT::Get::AllowUnauthenticated 1;',
+        )
+
     def add_repo(self, name, url, **kw):
         gpg_url = kw.pop('gpg_url', None)
         if gpg_url:
             self.add_repo_gpg_key(gpg_url)
+        else:
+            self.disable_signatures_check()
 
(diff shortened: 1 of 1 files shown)


dachary:
@alfredodeza hopefully this is in line with what you recommended during the discussions of https://github.com/ceph/ceph-deploy/pull/362 and https://github.com/ceph/ceph-deploy/pull/363. What do you think ?

alfredodeza:
@dachary we need to update the docs on contributing, but ceph-deploy has been (for a little while) using ticket identifiers on commit messages. This is prepared by a prepare-commit-msg script inside the `.git/hooks/` dir. The script is this https://gist.github.com/alfredodeza/6d62d99a95c9a7975fbe

If you look at pr #355 you will see that the requirement is to have the branch be the same as the ticket. BZ is used for Bugzilla tickets and RM for the redmine tracker.

Would you mind amending the commits to comply with that? (sorry for not having this contributing document ready to point there)

alfredodeza:
This should have a ticket in the tracker as well as in the description (and a link in the issue tracker back to this PR)

dachary:
@alfredodeza I'll do all that. Do you otherwise agree on the patch or does it go in the wrong direction ?

ceph_deploy/util/pkg_managers.py
```diff
@@ -235,10 +235,21 @@ def add_repo_gpg_key(self, url):
         cmd = ['apt-key', 'add', gpg_file]
         self._run(cmd)
 
+    def disable_signatures_check(self):
+        self.remote_conn.logger.info(
+            "Append APT::Get::AllowUnauthenticated 1 to /etc/apt/apt.conf"
+        )
+        self.remote_conn.remote_module.append_to_file(
+            '/etc/apt/apt.conf',
+            'APT::Get::AllowUnauthenticated 1;',
```
alfredodeza: this is a global setting and very unsafe to set it for an install and no set it back. Wouldn't the flag `--allow-unauthenticated` be better as a one time use case?

PR #367, opened 2015-10-19
[RM-13528] apt: no gpg key is installed allows unauthenticated packages

dachary:
Signed-off-by: Loic Dachary <ldachary@redhat.com>

### DIFF
--- a/ceph_deploy/util/pkg_managers.py
+++ b/ceph_deploy/util/pkg_managers.py
@@ -235,10 +235,21 @@ class Apt(PackageManager):
         cmd = ['apt-key', 'add', gpg_file]
         self._run(cmd)
 
+    def disable_signatures_check(self):
+        self.remote_conn.logger.info(
+            "Append APT::Get::AllowUnauthenticated 1 to /etc/apt/apt.conf"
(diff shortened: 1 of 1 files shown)


dachary:
> this is a global setting and very unsafe to set it for an install and no set it back. Wouldn't the flag --allow-unauthenticated be better as a one time use case?

That would be a flag to **ceph-deploy install** ?

alfredodeza:
`apt-get` allows a `--allow-unauthenticated` flag to do an install of a package without checking the authenticity. This would be a better option I think than setting a system wide change that would be permanent.

See: https://help.ubuntu.com/community/SecureApt in "Validation of Release files and Packages" section.

dachary:
@alfredodeza something like https://github.com/ceph/ceph-deploy/pull/362/files#diff-129ed243ea17f372404615bf22419ed1R58 is what you advise (except it would be **--allow-unauthenticated** intsead of **--force-yes** ) ?

alfredodeza:
Something like that yes, where 'extra flags' would mean to use the `--allow-unauthenticated`

PR #372, opened 2015-11-23
[RM-13585] Add timestamp to console and log output

mazebuhu:
An Example of the output here:
```
ceph-deploy gatherkeys mon1
[2015-11-23 11:55:09.049981][ceph_deploy.conf][DEBUG ] found configuration file at: /home/vagrant/.cephdeploy.conf
[2015-11-23 11:55:09.050858][ceph_deploy.cli][INFO  ] Invoked (1.5.28): /home/vagrant/bin/ceph-deploy gatherkeys mon1
[2015-11-23 11:55:09.051282][ceph_deploy.cli][INFO  ] ceph-deploy options:
[2015-11-23 11:55:09.051730][ceph_deploy.cli][INFO  ]  username                      : None
[2015-11-23 11:55:09.052110][ceph_deploy.cli][INFO  ]  verbose                       : False
```

I'm happy to discuss my proposed solution. 

For tracker see: http://tracker.ceph.com/issues/13585

### DIFF
--- a/ceph_deploy/util/log.py
+++ b/ceph_deploy/util/log.py
@@ -17,7 +17,7 @@ RESET_SEQ = "\033[0m"
 COLOR_SEQ = "\033[1;%dm"
 BOLD_SEQ = "\033[1m"
 
-BASE_COLOR_FORMAT = "[%(timestamp)s][$BOLD%(name)s$RESET][%(color_levelname)-17s] %(message)s"
+BASE_COLOR_FORMAT = "[$BOLD%(name)s$RESET][%(color_levelname)-17s] %(message)s"
 BASE_FORMAT = "[%(timestamp)s][%(name)s][%(levelname)-6s] %(message)s"
 
 
--- a/ceph_deploy/util/log.py
+++ b/ceph_deploy/util/log.py
@@ -1,5 +1,6 @@
 import logging
 import sys
+import datetime
 
 BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE = range(8)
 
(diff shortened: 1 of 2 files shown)


alfredodeza:
this looks OK but I think it would be better suited for the file-based logger (vs. both). Because the console is already quite cramped with the information.

mazebuhu:
yes you are right. It is too much if the timestamp is also displayed in the console output. Therefore I removed it there.

ceph_deploy/util/log.py
```diff
@@ -53,6 +54,7 @@ def format(self, record):
         truncated_level = record.levelname[:6]
         levelname_color = COLOR_SEQ % (30 + COLORS[levelname]) + truncated_level + RESET_SEQ
         record.color_levelname = levelname_color
+        record.timestamp = datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S.%f')
```
alfredodeza: Isn't this only for colored output formatting that goes only to the terminal/stderr? I guess it doesn't hurt to leave it there

mazebuhu:
@alfredodeza Maybe I was on the wrong way here in simply adding the a timestamp in the colored output formatting of the console.

I think a better implementation is to create a new *FILE_FORMAT* which adds the `%(asctime)s` in front of the *BASE_FORMAT* add setting it as format in the `cli.py` for the file logger. 

So we are sure that the timestamp get's only outputted to the log file and not also to a console which doesn't support color.

alfredodeza:
@mazebuhu agreed!

mazebuhu:
@alfredodeza ok pushed changes.

PR #375, opened 2015-12-10
[RM-14049] deal with systemd when stopping a mon otherwise raise an error

alfredodeza:
Issue tracker http://tracker.ceph.com/issues/14049
Signed-off-by: Alfredo Deza <adeza@redhat.com>

Successful output looks like this for a systemd box:

    ...
    [ceph_deploy.mon][DEBUG ] Removing mon from node4
    [node4][DEBUG ] connection detected need for sudo
    [node4][DEBUG ] connected to host: node4
    [node4][DEBUG ] detect platform information from remote host
    [node4][DEBUG ] detect machine type
    [node4][DEBUG ] find the location of an executable
    [node4][DEBUG ] get remote short hostname
    [node4][INFO  ] polling the daemon to verify it stopped
    [node4][INFO  ] Running command: sudo service ceph status mon.node4
    [node4][INFO  ] Running command: sudo mkdir -p /var/lib/ceph/mon-removed
    [node4][DEBUG ] move old monitor data

### DIFF
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -7,7 +7,7 @@ import time
 from ceph_deploy import conf, exc, admin
 from ceph_deploy.cliutil import priority
 from ceph_deploy.util.help_formatters import ToggleRawTextHelpFormatter
-from ceph_deploy.util import paths, net, files, packages
+from ceph_deploy.util import paths, net, files, packages, system
 from ceph_deploy.lib import remoto
 from ceph_deploy.new import new_mon_keyring
 from ceph_deploy import hosts
@@ -344,6 +344,14 @@ def destroy_mon(conn, cluster, hostname):
                 'status',
                 'mon.{hostname}'.format(hostname=hostname),
             ]
+        elif system.is_systemd(conn):
+            status_args = [
+                'systemctl',
+                'stop',
+                'ceph-mon@{hostname}.service'.format(hostname=hostname),
+            ]
+        else:
+            raise RuntimeError('unsupported init system detected, cannot continue')
 
         while retries:
             conn.logger.info('polling the daemon to verify it stopped')


alfredodeza:
please test this

alfredodeza:
test this please

smithfarm:
@alfredodeza LGTM - do you care to squash those two commits together before I merge?

smithfarm:
@alfredodeza Also, over in `ceph/ceph.git` there is a convention of including the tracker issue number in the commit message like this: `Fixes: #14049` - I take it ceph-deploy uses a different convention?

alfredodeza:
@smithfarm https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst 

The convention is basically to have it prefixed in the commit message (as is on this PR)

smithfarm:
Hopefully I'm not being too pedantic. . . This commit https://github.com/ceph/ceph-deploy/commit/6bdbfe74813744476681074f5578737795d9b728 uses both conventions. I think the tracker issue then magically displays information on the commit that fixes it.

smithfarm:
Ah, but that tracker magic is unreliable - no commit info shown in http://tracker.ceph.com/issues/12999

smithfarm:
@alfredodeza Looking at that `if` statement, how confident are you that it won't trigger in the first two clauses (initctl, upstart) on a systemd system? Would it make sense to move the `is_systemd` clause to the top of the `if` ?

alfredodeza:
@smithfarm I am actually very confident, because if you look at both the tracker and bugzilla tickets:

http://tracker.ceph.com/issues/14049
https://bugzilla.redhat.com/show_bug.cgi?id=1282484

The bug is triggered because in a systemd server none of those `if/elif` conditions was met, therefore the `status_args` was never defined.

smithfarm:
@alfredodeza Yeah, I understand that on one or two systemd-based systems those stanzas are not triggered, but that can't be extrapolated to *all* systemd-based systems without further analysis. 

The conditionals are evaluating this: `conn.remote_module.path_exists(os.path.join(path, 'upstart'))` - do you know what path that is, exactly? I unsuccessfully tried to figure it out last night.

smithfarm:
Aha! There it was, staring me in the face: https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/util/paths/mon.py#L23

I looked on a systemd Hammer box and sure enough there is a `systemd` file there:

```
teuthida-1:/var/lib/ceph/mon/ceph-teuthida-1 # ls
done  keyring  store.db  systemd
```

PR #377, opened 2015-12-15
[RM-13909] bump requirement for remoto

alfredodeza:
fixes http://tracker.ceph.com/issues/13909

With the following commit:
https://github.com/alfredodeza/remoto/commit/f9c8e87a8e3fc48bd122c86953ec87ac1be13a10

remoto 0.0.26 is now released

/cc @ktdreyer

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ if os.environ.get('CEPH_DEPLOY_NO_VENDOR'):
     clean_vendor('remoto')
 else:
     vendorize([
-        ('remoto', '0.0.25', ['python', 'vendor.py']),
+        ('remoto', '0.0.26', ['python', 'vendor.py']),
     ])
 
 


ktdreyer:
Can we stop vendoring remoto, now that the [execnet change](https://bitbucket.org/hpk42/execnet/pull-requests/12/dont-call-sys-when-interpreter-is-shutting/diff) we need is present in execnet 1.3.0 ?

alfredodeza:
we can't stop vendoring them because they are not present in the distros that we support, unless we want to commit to packaging execnet and remoto for ceph-deploy?

ktdreyer:
Oh, ok, so packaging is the only remaining blocker.

alfredodeza:
@ktdreyer yes. If we stop vendoring (we can if we were using it as pure python only) then we would need to require both remoto and execnet to be available for others to consume, like ceph-deploy

ktdreyer:
Thanks, I get it now.

PR #383, opened 2016-01-20
[RM-13847] Using full paths to executables

mazebuhu:
Refactored out some direct executable calls in "osd.py". 

Should we do that for all executable calls in ceph-deploy?

For tracker see: http://tracker.ceph.com/issues/13847

### DIFF
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -107,8 +107,9 @@ def osd_status_check(conn, cluster):
     Note how the booleans are actually strings, so we need to take that into
     account and fix it before returning the dictionary. Issue #8108
     """
+    ceph_executable = system.executable_path(conn, 'ceph')
     command = [
-        'ceph',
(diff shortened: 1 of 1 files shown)


alfredodeza:
ok to test

ceph_deploy/osd.py
```diff
@@ -424,10 +427,11 @@ def disk_zap(args):
         if distro.normalized_name.startswith(('centos', 'red')):
             LOG.info('calling partx on zapped device %s', disk)
             LOG.info('re-reading known partitions will display errors')
+            partx_executables = system.executable_path(distro.conn, 'partx')
```
alfredodeza: why is this `partx_executables` (plural) ?

mazebuhu: It's a typo. Sorry.

mazebuhu:
Ok I fixed the typo.

PR #384, opened 2016-02-02
ceph-deploy: Disable timeout for pkg install/remove procedures

borix:
Do not timeout too early, here. If we use the pkg command to install
e.g. ceph-selinux or some other package with long post script we can
easily timeout in 5 minutes, disabling the timeout completely for the
pkg install procedure makes this much more safe.

The same holds for pkg removal.

Signed-off-by: Boris Ranto <branto@redhat.com>

### DIFF
--- a/ceph_deploy/pkg.py
+++ b/ceph_deploy/pkg.py
@@ -17,6 +17,12 @@ def install(args):
         )
         rlogger = logging.getLogger(hostname)
         rlogger.info('installing packages on %s' % hostname)
+        # Do not timeout on package install. If you we this command to install
+        # e.g. ceph-selinux or some other package with long post script we can
+        # easily timeout in the 5 minutes that we use as a default timeout,
+        # turning off the timeout completely for the time we run the command
+        # should make this much more safe.
+        distro.conn.global_timeout = None
         distro.packager.install(packages)
         distro.conn.exit()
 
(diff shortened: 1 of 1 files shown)


alfredodeza:
ok to test

ceph_deploy/pkg.py
```diff
@@ -17,7 +17,15 @@ def install(args):
...
+        # Do not timeout on package install. If you we this command to install
+        # e.g. ceph-selinux or some other package with long post script we can
+        # easily timeout in the 5 minutes that we use as a default timeout,
+        # turning off the timeout completely for the time we run the command
+        # should make this much more safe.
+        backup_timeout = distro.conn.global_timeout
+        distro.conn.global_timeout = None
         distro.packager.install(packages)
+        distro.conn.global_timeout = backup_timeout
         distro.conn.exit()
```
alfredodeza: since the connection ends here, there is no need to backup the global timeout to reset it.

'global' in this context means that it will be used for all actions during a connection. `distro.conn.process.run` has a more granular timeout, which defaults to "wait for ever", the `global` takes precedence and that is why your changeset is correct in changing that (vs. the more granular one).

borix:
The new patch should cover Alfredo's comment.

btw: When looking at the code, while timeout usually takes precedence over global_timeout, it won't do so if timeout is 0 or None because of this line in remoto code:

timeout = timeout or global_timeout

I'd probably go as far as consider it a bug (or design flaw) in remoto/underlying libraries.

ktdreyer:
@alfredodeza can you merge this please?

alfredodeza:
@borix @ktdreyer there is one last thing here, the commits need to be reworded:

https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#commits

borix:
Rebased with the updated commit message.

PR #387, opened 2016-02-11
Fix mon install suse

oms4suse:
At the moment when testing with:

    https://build.opensuse.org/package/show/filesystems:ceph:jewel/ceph

mon create fails, these 2 patches fix this issue.

### DIFF
--- a/ceph_deploy/hosts/suse/mon/create.py
+++ b/ceph_deploy/hosts/suse/mon/create.py
@@ -1,4 +1,5 @@
 from ceph_deploy.hosts import common
+from ceph_deploy.util import system
 from ceph_deploy.lib import remoto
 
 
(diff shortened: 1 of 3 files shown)


alfredodeza:
@oms4suse would you mind creating an issue for this and follow the contributing guidelines for commits? They are described here: https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#contributing-to-ceph-deploy

Thanks

oms4suse:
@alfredodeza hope this is better now.

alfredodeza:
test this please

ceph_deploy/hosts/suse/mon/create.py
```diff
@@ -6,14 +7,50 @@ def create(distro, args, monitor_keyring):
...
+                '/etc/ceph/{cluster}.conf'.format(cluster=args.cluster),
+                'start',
+                'mon.{hostname}'.format(hostname=hostname)
+            ],
+            timeout=7,
+        )
+
+        system.enable_service(distro.conn)
+    elif distro.init == 'systemd':
+       # enable ceph target for this host (in case it isn't already enabled)
```
alfredodeza: is it ok to keep enabling this if it has already been enabled? would you mind clarifying/verifying this?

alfredodeza:
There are failures with this PR that you will also need to fix. Please make sure you are running the unit tests when submitting a pull request

ceph_deploy/hosts/suse/mon/create.py
oms4suse: Sure. Have been in conference all last week, will get to this again after Thursday.

oms4suse: I could not find an explicit documentation stating that you can enable the same service N times. But I did find that it will create symlinks to persist the information of enabling a service, and will not start a service when enabling.

We have an exception to this when the parameters of the parametrised service is different but this is as would be expected.

oms4suse:
@alfredodeza I think these tests are now fixed.

PR #388, opened 2016-03-08
[RM-15016] Python 3 Compatibility

BlaXpirit:
http://tracker.ceph.com/issues/15016

These changes allow Ceph to run on Python 3.x while keeping compatibility with Python 2.x

All tests from the test suite (tox) pass on Python 3.4.2 and Python 2.7.9 (Debian GNU/Linux 8.3 (jessie) Linux 4.3.0-0.bpo.1-amd64).
`2033 passed, 16 skipped in 4.79 seconds` // `py34: commands succeeded`
(actually, on Python 2.7.9 one test (*test_simple*) fails but that's [an unrelated problem](http://tracker.ceph.com/issues/15018))

Rudimentary usage (`ceph-deploy admin admin-node`) works.

~~Work on this is postponed until [GSOC](https://ceph.com/gsoc2016/) starts, unless the maintainers find this change important to include sooner.~~

TODO:
- Refine the code
    - [x] More meaningful separation into commits and commit messages
    - [x] Maybe some `#noqa` can be dropped
        - nope
    - [x] Spacing of imports
- Respond to feedback
    - [x] Should [`six`](https://pythonhosted.org/six/) be used or avoided?
        - No `six`
- Check for possible unaccounted problems
    - [x] Configs and filenames are stored in native strings (bytes in py2, unicode in py3) &ndash; is that OK?
- [x] Check if it works on Python 2.6 and what other versions of 3.x it can run on
    - 2.6, 2.7, 3.3, 3.4, 3.5
- [x] Decide whether Python 2 <-> Python 3 support is viable (different Python version on remote machine)
    - nope
- [x] Required changes to *remoto* - wait for merge + release
- [x] Integration testing
    - [x] Python 3 isn't installable on CentOS 7 - what to do?
        - epel-release repository

/cc @jdurgin


alfredodeza:
ok to test

ceph_deploy/osd.py
```diff
@@ -674,6 +675,7 @@ def make(parser):
     parser.description = sub_command_help
 
     osd_parser = parser.add_subparsers(dest='subcommand')
+    osd_parser.required = True
```
alfredodeza: I am not sure why this is needed, mind explaining this a bit?

ceph_deploy/osd.py
```diff
@@ -593,7 +594,7 @@ def print_osd(logger, hostname, osd_path, json_blob, metadata, journal=None):
     logger.info('%-14s %s' % ('Reweight', json_blob.get('reweight')))
     if journal:
         logger.info('Journal: %s' % journal)
-    for k, v in metadata.items():
+    for k, v in list(metadata.items()):
```
BlaXpirit: Yes, the automatically-added `list` is not needed. I will fix things like these in the future.

ceph_deploy/tests/util.py
```diff
@@ -24,5 +26,12 @@ class Empty(object):
     instantiation.
     """
     def __init__(self, **kw):
-        for k, v in kw.items():
+        for k, v in list(kw.items()):
             setattr(self, k, v)
+
+
+def assert_too_few_arguments(err):
```
alfredodeza: I am surprised for this addition, would you explain this as well? py.test (the test runner we use) does a great job at managing assertions and comparisons

ceph_deploy/osd.py
BlaXpirit: Python 3.4 does not give an error when an incomplete command is passed, e.g. `ceph-deploy mon`, making it crash with an exception later down the line.

More info: http://bugs.python.org/issue9253#msg186387

Minimal example to compare behavior of Pythons:
```python
import argparse
parser = argparse.ArgumentParser()
parser.add_subparsers()
parser.parse_args(args=[])
```

ceph_deploy/tests/util.py
BlaXpirit: I didn't want to repeat the clunky double check. It appears in many places in the code base. So I just put this piece of code in a function.

Now, the reason that this new check is needed in the first place  is simply that *argparse* now has different error messages, and the tests failed expecting the ones from Python 2.x.

BlaXpirit:
Something went wrong, I must have omitted some final changes to the code...

BlaXpirit:
I realize that I made in a dependency. So I reinitialized everything and detected another change. Naturally, it was not submitted here.

```diff
diff -U6 -r -x '*.pyc' ceph-deploy/ceph_deploy/lib/vendor/remoto/process.py ceph-deploy_/ceph_deploy/lib/vendor/remoto/process.py
--- ceph-deploy/ceph_deploy/lib/vendor/remoto/process.py        2016-03-09 00:48:09.448870343 +0200
+++ ceph-deploy_/ceph_deploy/lib/vendor/remoto/process.py       2016-03-08 23:18:45.885113354 +0200
@@ -136,14 +136,15 @@
 def _remote_check(channel, cmd, **kw):
     import subprocess
 
     process = subprocess.Popen(
         cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kw
     )
-    stdout = [line.strip('\n') for line in process.stdout.readlines()]
-    stderr = [line.strip('\n') for line in process.stderr.readlines()]
+    line = None # remoto/execnet bug
+    stdout = [line.strip(b'\n') for line in process.stdout.readlines()]
+    stderr = [line.strip(b'\n') for line in process.stderr.readlines()]
     channel.send((stdout, stderr, process.wait()))
 
 
 def check(conn, command, exit=False, timeout=None, **kw):
     """
     Execute a remote command with ``subprocess.Popen`` but report back the
```

And no, I don't see a way around this change. One is obviously a bug because `'\n'.join([b'list', b'of', b'bytestrings'])` is definitely no good.
And the `line = None` part prevents another obscure bug that you may see in the logs:
> the use of non-builtin globals isn't supported
There is some weird function serialization going on, and it thinks that `line` is a global because it's used in the function without previously being assigned.

alfredodeza:
@BlaXpirit the remoto change does not belong to this repo :) If you check carefully you will realize it actually is a different git checkout.

remoto lives here https://github.com/alfredodeza/remoto

And when calling setup.py in ceph-deploy it "pulls" in remoto.

BlaXpirit:
@alfredodeza I know.

> I realized that I made a change in a dependency.

Didn't see that it was your repository, though. So submitting a fix should not be a problem.

jhdulaney:
What is the status of this?  I'd like to work this into Fedora.

BlaXpirit:
Welp, no matter what I do the release of remoto is forced to 0.0.27. How does that even work?

alfredodeza:
@BlaXpirit `0.0.28` is out for remoto, and have updated ceph-deploy (in master) to request it

BlaXpirit:
@alfredodeza Alright, thanks. The tests on Jenkins passed even earlier, but then Jenkins itself errors out (seen in Console Output).

BlaXpirit:
Whew, finally the tests passed on Jenkins. There were all kinds of bugs with GitHub partial's downtime.

So I have rearranged this whole branch, grouping the changes into commits by topic (preserved the original sequence here: https://github.com/ceph/ceph-deploy/compare/master...BlaXpirit:RM-15016-WIP)

I decided to drop *six* in the end but it's not too late to put it back in. Here's the difference with/without *six*: https://github.com/BlaXpirit/ceph-deploy/commit/cb6a0272e194ffe4f44ecaa390d8194feba30879. Sure, it's nicer with it but it's still a 3rd party dependency, which *ceph-deploy* kinda doesn't have right now.

@jdurgin Now I consider this complete. Ready for review and next steps.

jdurgin:
Nice work! Grouping the changes like that made it a lot easier to review.
six wasn't avoiding much code, so I'm happy you ended up not using it here.

The changes all looked good to me, so I pushed them to a branch (wip-15016) and scheduled an integration test run against it: http://pulpito.ceph.com/joshd-2016-05-24_00:35:48-ceph-deploy-jewel---basic-mira/ which just passed. Looks ready to merge to me.

@alfredodeza any comments on this?

BlaXpirit:
@jdurgin Are you sure these tests checked the code running on Python 3?

And another thing, it would be ideal to add more Python versions to the unittest runner, but currently I think Python 3.3 and Python 3.5 are not available through Jenkins.

jdurgin:
@BlaXpirit I'm sure they did not use python 3 :) They just checked for regressions for python 2.

To get the integration tests working with python 3 I think we'll need to update https://github.com/ceph/ceph-qa-suite/blob/master/tasks/ceph_deploy.py#L186 and possibly the venv created by ceph-deploy's bootstrap script to explicitly specify the python version.

Then we can add python 3 as a facet to ceph-qa-suite/suites/ceph-deploy. We can go over details on how this works more on irc, tonight or tomorrow.

BlaXpirit:
Everything here is ready.

This should be merged along with https://github.com/ceph/ceph-qa-suite/pull/1058 - also check there to see test runs with these changes

PR #389, opened 2016-03-11
osd,disk: add --bluestore flag

liewegas:
This will pass --bluestore on to ceph-disk during the prepare/create
step.

Tested on my local jessie box.  Works, except that ceph-detect-init is wrong,
but that's a different issue.


vasukulkarni:
:+1:  thanks for making this simple to deploy

vasukulkarni:
I will be adding some tests on this options as well

PR #392, opened 2016-04-07
Fix wrong log level in ceph deploy log

jordan41177:
In command of ceph-deploy, the subcommand's log is in wrong log level.
Before fix
image

After fix
image


jordan41177:
http://tracker.ceph.com/issues/15410

alfredodeza:
Looks good. A couple of things though:

* the "add comma" commit should probably be squashed onto the previous commit
* commits need to be prefixed with the tracker's ticket ID (see https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#commits)

alfredodeza:
OK to test

jordan41177:
Hi @alfredodeza 

I fixed it. Thank you

alfredodeza:
It seems this wouldn't work for older versions of Ceph like Hammer. 

    ceph-disk: error: unrecognized arguments: --log-stdout
    vagrant@vagrant:~$ ceph --version
    ceph version 0.94.6 (e832001feaf8c176593e0325c8298e3f16dfb403)

ceph-deploy supports most Ceph versions including hammer and this would break when not using the latest versions. Can't think of anything in ceph-deploy that would allow you to conditionally support this for newer versions only.

jordan41177:
Hi @alfredodeza  
I will check the tool 'ceph-disk' in other version.
I tested this in Jewel, and I would try other version after weekend.

Thank you

alfredodeza:
ceph-disk is no longer part of ceph

PR #395, opened 2016-04-13
[RM-15018] Remove broken 'mon create' unittest

BlaXpirit:
Details at http://tracker.ceph.com/issues/15018

@alfredodeza says
> I am inclined to say this test is not very good and should be removed

But then I realized that the `new` call is actually tested properly, and it seems worth keeping.

### DIFF
--- a/ceph_deploy/tests/test_cli_mon.py
+++ b/ceph_deploy/tests/test_cli_mon.py
@@ -1,9 +1,7 @@
-import argparse
-import collections
 import subprocess
 
 import pytest
-from mock import Mock, patch, NonCallableMock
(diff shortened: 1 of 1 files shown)


ceph_deploy/tests/test_cli_mon.py
```diff
@@ -30,43 +28,22 @@ def make_fake_connection(platform_information=None):
     return get_connection
 
 
-def test_simple(tmpdir, capsys):
+def test_new(tmpdir, capsys):
```
alfredodeza: thank you

alfredodeza:
OK to test

PR #396, opened 2016-04-21
default to jewel

liewegas:


dmick:
can't we make this look it up?  what a drag to have to edit the string for every release

vasukulkarni:
I dont think its that easy, only after the release mail is sent this has to be changed and it would be easier to change here then someplace else. Current logic looks good to me.

alfredodeza:
@dmick "looking it up" means having a *rock solid* place where to put this (we don't have one that I feel confident enough), and it also assumes that ceph-deploy users will be able to hit it in any/all environments. 

Maybe a combination of solutions should be discussed and find some middle ground. The current approach will not help people with older versions of ceph-deploy :( Ans this merge will not do anything for them until I cut a release and they upgrade.

dmick:
it only needs to be as solid as the release repos themselves, and even a symlinked "stable -> <jeweltoday>" (or whatever name) seems better to me.

I just hate reiterating authoritative information.

PR #402, opened 2016-06-14
Fix rsync username when using local_mirror

gaowanlong:
We should use the configured username for rsync to connect
to the remote host.

Signed-off-by: Wanlong Gao <wanlong.gao@easystack.cn>


alfredodeza:
OK to test

PR #403, opened 2016-06-15
Do not use sudo in rsync

gaowanlong:
Got the error with:

	[node-1][INFO  ] installing Ceph on node-1
	sudo: sorry, you must have a tty to run sudo

Seems that the rsync do not have a tty to run sudo, so
do not use sudo here and just ensure that the user has
permission to sync the packages to the right location.

Signed-off-by: Wanlong Gao <wanlong.gao@easystack.cn>

### DIFF
--- a/ceph_deploy/install.py
+++ b/ceph_deploy/install.py
@@ -149,7 +149,7 @@ def install(args):
         if args.local_mirror:
             if args.username:
                 hostname = "%s@%s" % (args.username, hostname)
-            remoto.rsync(hostname, args.local_mirror, '/opt/ceph-deploy/repo', distro.conn.logger, sudo=True)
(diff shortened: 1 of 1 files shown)


alfredodeza:
OK to test

alfredodeza:
@gaowanlong this is assuming the user has permissions to write to that path.

The tty error is solved with configuration of `/etc/sudoers` with `!requiretty`

An even better approach would be to modify remoto to auto detect the need for sudo (which it can do but it doesn't for rsync):

It would entail updating this call to `Connection` https://github.com/alfredodeza/remoto/blob/master/remoto/file_sync.py#L37 so that it passes `detect_sudo=True` which will try to see if sudo is really needed: https://github.com/alfredodeza/remoto/blob/master/remoto/connection.py#L42-L50

gaowanlong:
@alfredodeza Good to know, thank you for you explanation, then this is not a issue any more.

PR #407, opened 2016-06-24
Handle ubuntu package split for ceph-osd and ceph-mon

vasukulkarni:
ceph is now split into ceph-mon and ceph-osd, this might actually break hammer installs if the packages are not split there, needs some good testing.

### DIFF
--- a/ceph_deploy/hosts/debian/install.py
+++ b/ceph_deploy/hosts/debian/install.py
@@ -4,17 +4,10 @@ except ImportError:
     from urlparse import urlparse
 
 from ceph_deploy.util.paths import gpg
-from ceph_deploy.hosts.common import map_components
-
-
-NON_SPLIT_COMPONENTS = ['ceph-osd', 'ceph-mon']
 
 
 def install(distro, version_kind, version, adjust_repos, **kw):
-    packages = map_components(
-        NON_SPLIT_COMPONENTS,
-        kw.pop('components', [])
-    )
+    packages = kw.pop('components', [])
     codename = distro.codename
     machine = distro.machine_type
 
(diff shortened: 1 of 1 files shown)


vasukulkarni:
Also depends on https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1596063

alfredodeza:
This PR needs to follow the conventions for the project. Please see https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#contributing-to-ceph-deploy

alfredodeza:
This change *will* break Hammer installations. There is no package split there. We should either not care about Hammer or make changes to accommodate Hammer installs and special case the component names that need to be used for installation.

I prefer the latter :)

vasukulkarni:
@alfredodeza sure, I will fix the conventions required, Regarding hammer I am not sure how CentOS is behaving there, I am going to run some tests in qa-suite today, Since Hammer is out for some time, it might be ok to cut the release that supports only jewel+ and document for hammer to use specific version of ceph-deploy or we could also fix package split in hammer :) , we can very well handle the hammer vs jewel install in ceph-deploy but the repo has to be queried for that? Possibly chacra/cli/etc?

vasukulkarni:
@alfredodeza surprising it worked for hammer as well
hammer: http://pulpito.ceph.com/vasu-2016-06-27_12:13:37-ceph-deploy-hammer---basic-vps/
jewel: http://pulpito.ceph.com/vasu-2016-06-27_11:56:48-ceph-deploy-jewel---basic-vps/

i think the package split was handled in hammer probably

vasukulkarni:
@alfredodeza changed the commit to include bz issue and the tests pass for hammer and jewel.

PR #410, opened 2016-07-09
modify sysvinit if condition

ieth0:
If ceph packages aren't installed on the host with systemd preinstalled, script will return sysvinit instead of systemd which is invalid.


alfredodeza:
OK to test

PR #412, opened 2016-07-19
Cleanup unneeded requirements

hguemar:
python-distribute conflicts with setuptools and is deprecated.
Moreover we don't run unit test at build time

### DIFF
--- a/ceph-deploy.spec
+++ b/ceph-deploy.spec
@@ -20,19 +20,14 @@ URL:            http://ceph.com/
 Source0:        %{name}-%{version}.tar.bz2
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  python-devel
-BuildRequires:  python-distribute
 BuildRequires:  python-setuptools
-BuildRequires:  python-virtualenv
 BuildRequires:  python-mock
-BuildRequires:  python-tox
 %if 0%{?suse_version}
 BuildRequires:  python-pytest
 %else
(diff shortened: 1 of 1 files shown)


alfredodeza:
@hguemar could you please create an issue in the tracker for this fix and try to go over the guide at https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#contributing-to-ceph-deploy 

Some of your changes would break for Python 2.6 (e.g. it removes the python-argparse requirement). This PR also removes python-tox but doesn't remove python-mock. I think that the removal of those test dependencies is good, but can you remove *all* of the test dependencies?

ktdreyer:
LGTM. We can always add test deps back if/when we can ever get the tests to pass during rpmbuild.

hguemar:
@alfredodeza I'll rework that patch as per your suggestion.

alfredodeza:
@hguemar ping?

mcronce:
Bump...this would be real nice to have, as it's not possible to install the Luminous version of ceph-deploy on Fedora - it's provided by another package now:

```
$ rpm -qf /usr/lib64/python2.7/argparse.py
python2-libs-2.7.14-2.fc26.x86_64
```

alfredodeza:
@mcronce would you be willing to take over this PR and update @hguemar commits to follow the contributing guidelines? This needs to be rebased as well. I don't mind a new PR and crediting the initial work

PR #418, opened 2016-08-24
[RM-16443] gatherkeys: try getting an existing key without caps

alfredodeza:
In case the key already exists but the caps do not match

See: http://tracker.ceph.com/issues/16443


vasukulkarni:
This fixes the issue for mon create initial failure, sadly the ceph-deploy hammer test branch is behind and is not accurate since its missing backports, there are 2 other issues to fix and I will raise a tracker for that.

PR #422, opened 2016-10-13
[RM-17567] Start mon by systemd/sysvint/upstart after adding mon

adshao:


vasukulkarni:
All the repetitive code for various OS'es is not required, the deleting part in common is fine and probably calling a function that does start in generic way for all OS'es should be good way to avoid repetition.

adshao:
hi @vasukulkarni , so we need to remove the mon folder (including create.py) from all OS'es?

alfredodeza:
This looks OK to me. @vasukulkarni can you double check please?

alfredodeza:
OK to test

vasukulkarni:
@alfredodeza the new changes does look good to me  :+1:  , I believe jenkins failure is due to py34

PR #423, opened 2016-11-04
osd: Re-add ability to pick ext4 as an fs type

iain-buclaw-sociomantic:
However it is explicitly undocumented in the help output, and requires
adding --yes-i-really-mean-it in order to use it.

Signed-off-by: Iain Buclaw <iain.buclaw@sociomantic.com>

### DIFF
--- a/ceph_deploy/osd.py
+++ b/ceph_deploy/osd.py
@@ -316,6 +316,13 @@ def prepare(args, cfg, activate_prepared_disk):
             if args.bluestore:
                 storetype = 'bluestore'
 
+            if args.fs_type == 'ext4' and not args.yes_i_really_mean_it:
+                LOG.error('Warning! The ceph-osd daemon will refuse to start or may have difficult '
+                          'to diagnose errors if it cannot safely store long object names on ext4.')
+                LOG.error('Add --yes-i-really-mean-it if you are certain.')
+                errors += 1
+                continue
+
             prepare_disk(
                 distro.conn,
                 cluster=args.cluster,
@@ -678,6 +685,7 @@ def make(parser):
         '--fs-type',
         metavar='FS_TYPE',
         choices=['xfs',
+                 'ext4',
                  'btrfs'
                  ],
         default='xfs',
@@ -698,6 +706,11 @@ def make(parser):
         '--bluestore',
         action='store_true', default=None,
         help='bluestore objectstore',
+        )
+    osd_create.add_argument(
+        '--yes-i-really-mean-it',
+        action='store_true', default=None,
+        help='enable experimental or deprecated features',
         )
     osd_create.add_argument(
         'disk',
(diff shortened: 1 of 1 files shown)


iain-buclaw-sociomantic:
My argument for putting this back in is that I'm only using librados directly for interacting with ceph (no RGW or other frontend solutions).

I have benchmarked cluster speed using our typical load of reads/writes, with object sizes on average are around 4kb.  Only btrfs and ext4 managed to get up to tens of thousands of operations a second, whereas xfs only managed to peak at 1000 initially (and went downhill from there).

Using btrfs in a modest sized production environment has proved that btrfs weighs more negatively on continuous maintenance over the positive throughput and speed.

Meanwhile, in our pre-Ceph storage system, we have been using ext4 for years with no problems, so I'd like to continue using it for the moment while still using ceph-deploy to manage the cluster.

Yes, I know what I'm doing when I say I want ext4.  And yes, I will be eagerly waiting for Bluestore to become stable. :-)

vasukulkarni:
@liewegas ok with this?

liewegas:
Yep!

ceph_deploy/osd.py
```diff
@@ -742,6 +751,11 @@ def make(parser):
         help='bluestore objectstore',
         )
     osd_prepare.add_argument(
+        '--yes-i-really-mean-it',
```
vasukulkarni: sorry i had submitted this before but was in pending state

iain-buclaw-sociomantic: Done, and rebased.

vasukulkarni:
LGTM, @alfredodeza ping

alfredodeza:
I am going to push back on this. `ceph-deploy` is meant to be a one-stop-shop for easy ceph deployments with known (and sometimes forced) defaults. Adding a flag to go around a well known default (even with the new warning) looks like special purpose to me.

Anything that deviates from "this should work fine for everyone that wants to try ceph" shouldn't get in, and was the reason why we decided to fully remove the option of ext4.

This reasoning is also fairly well explained in the README: https://github.com/ceph/ceph-deploy#what-this-tool-is-not

@iain-buclaw-sociomantic thank you for the PR. But at this point I highly recommend using a deployment system that allows a different configuration, like ceph-ansible.

iain-buclaw-sociomantic:
@alfredodeza - Thanks, we use puppet here, however puppet-ceph is in a shamble state.  It is very difficult to ensure a strict order of operations happen across many nodes, and that is where writing a soft wrapper around ceph-deploy came in.

alfredodeza:
@iain-buclaw-sociomantic understood. I believe ceph-deploy is used in other Puppet-based implementations for certain things too. But deviating from the goals of what ceph-deploy is will make it easier for one-off/single-purpose use cases to make it in which we have tried to avoid for a while.

As long as changes are generic and can apply to the most common use case(s) then I don't have a problem getting them in.

PR #424, opened 2016-11-08
[RM-17823] gatherkeys: add missing argument to log call

alfredodeza:
See: http://tracker.ceph.com/issues/17823


vasukulkarni:
http://i.imgur.com/xlkybSn.gif

PR #425, opened 2016-11-24
Rm 18035,rm 18034,rm 18033,rm 18032,rm 18031,rm 18029,rm 18028

oms4suse:
This is a set of downstream work to deal with issues:

1. #18028 rgw cannot define custom port
2. #18029 rgw does not provide a way to delete configured services.
3. #18031 rgw does not provide a way to list rgw services.
4. #18032 rgw option does not specify the dns hostname in the config file.
5. #18033 rgw does not specify the front end in the config file.
6. #18034 rgw takes Czech characters as service name then fails.
7. #18035 rgw should not default the name when users may have a good reason to specify non defaulted name.

### DIFF
--- a/ceph_deploy/rgw.py
+++ b/ceph_deploy/rgw.py
@@ -152,6 +152,10 @@ def rgw_create(args):
     # Update the config file
     changed_cfg = False
     for hostname, name in args.rgw:
+        if not name.startswith('rgw.'):
+            msg = "rgw name '%s' does not start with 'rgw.'" % (name)
+            LOG.error(msg)
+            raise RuntimeError(msg)
         enitity = 'client.{name}'.format(name=name)
         port = 7480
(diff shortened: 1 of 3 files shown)


vasukulkarni:
@oms4suse 
1) needs doc update as a dependent pr since its adding new features: http://docs.ceph.com/docs/master/rados/deployment/ 
2) we should add rgw test to ceph-deploy suite for config/list/delete
3) there is one merge commit in the pr and since its not a merge here, it will need better description
4) docstring for new functions

alfredodeza:
The commits here are not formatted according to the contributing guidelines. Please make sure that you are following https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst

alfredodeza:
jenkins test this please

PR #429, opened 2017-02-03
Allow unsigned deb packages install from mirror locations

ganeshmaharaj:
In a test deployment, hosting ceph packages in a local mirror and
installing them across the cluster will be ideal. Current ceph-deploy
allows unauthenticated packages when install is used but not from a
mirror location. This change helps address that usecase.

Signed-off-by: Ganesh Mahalingam <ganesh.mahalingam@intel.com>

### DIFF
--- a/ceph_deploy/hosts/debian/install.py
+++ b/ceph_deploy/hosts/debian/install.py
@@ -76,6 +76,7 @@ def install(distro, version_kind, version, adjust_repos, **kw):
 
 def mirror_install(distro, repo_url, gpg_url, adjust_repos, **kw):
     packages = kw.pop('components', [])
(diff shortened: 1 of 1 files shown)


ganeshmaharaj:
I understand using '--dev' in this method is not its intended purpose. But i thought in this way of install it is something that can be extended to help with this. Please feel free to correct me if i am wrong in the usage or if there is a cleaner way of handling this change.

vasukulkarni:
LGTM, if you can post some run logs from the mirror would be nice.

ganeshmaharaj:
@vasukulkarni Thanks. Please find logs for a run with --repo-url --dev here. https://hastebin.com/ofuleluzur.md

PR #435, opened 2017-03-13
add mgr support (install, gatherkeys, daemon creation)

liewegas:


ceph_deploy/install.py
```diff
@@ -55,6 +56,7 @@ def detect_components(args, distro):
         'install_rgw': 'ceph-radosgw',
         'install_mds': 'ceph-mds',
         'install_mon': 'ceph-mon',
+        'install_mon': 'ceph-mgr',
```
alfredodeza: this key is duplicated, so it is overriding

vasukulkarni:
@alfredodeza i think we should cut the jewel branch before mgr changes, Not sure the code will move to jewel but looking at other bluestore changes i believe we could push the stable jewel version to pypi and also cut a branch here.

liewegas:
Tests now pass!

The gatherkeys changes should not effect jewel (or older) installs.  However, the install task now tries to install the ceph-mgr package, which doesn't exist for jewel.  I'm not sure how to handle that... suggestions?  Maybe a --no-mgr flag that can be passed for legacy versions?  Or separate it into a different install step and warn instead of error out if it can't be found?

liewegas:
(Either way, we should make sure that master still is able to install kraken and jewel!)

vasukulkarni:
I think the easier approach would be not to install mgr by default for all and then install only using --install-mgr flag which can be done easily with 'components' list.  It would require user to run one more command. 

Nicer way would be to detect 'jewel', 'kraken' etc but that can be done only after install or we could simplify that by using a flag in ~/.ceph_deploy.conf which was discussed during hammer time.

 Also I believe it should be fine to branch out for 'jewel' and modify tests to use jewel branch for jewel tests, I dont there there will be lot of back ports for that branch, but if we are taking bluestore changes for jewel then it doesn't make sense to branch out.

liewegas:
this seems reasonable for now; after luminous is out we can switch teh 
default.  although even then we'll still need a 
--no-install-mgr flag... so we may as well add that now too?

vasukulkarni:
Sure, I guess it makes sense to add the flag now.

liewegas:
Confirmed that install works with both jewel and master for rpm and deb.  I think this is ready to merge?

liewegas:
retest this please

vasukulkarni:
using 'ceph' metapackage was nice :+1:

vasukulkarni:
http://pulpito.ceph.com/vasu-2017-03-17_17:00:23-ceph-deploy-jewel---basic-vps/

liewegas:
Thanks, Vasu!

PR #440, opened 2017-05-19
remove py34 tests until fixed

vasukulkarni:
Remove py34 tests until it works 

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26, py27, py34, flake8
+envlist = py26, py27, flake8
 
 [testenv]
 deps=


alfredodeza:
there is no need to remove this, we could pass the flag to tox to ignore if the interpreter is not found, but I don't think we even need to do that, the fix for this is not that complicated and it is all related to the build system, not ceph-deploy

vasukulkarni:
@alfredodeza whats the plan? are you going to fix the build system then? This PR was to avoid ignoring real errors since its been failing for quite some time now.

alfredodeza:
@vasukulkarni the PR wasn't avoiding running tests that are passing, and it would mean it would not be possible to run anywhere (e.g. I wouldn't be able to run those locally). 

The build system needs to be updated, this is not the top priority right now, I would be happy to review an update though! The changes would mean adding a 'python3' label to mita machines that are really xenial (github.com/ceph/mita ) and then update the ceph-build.git repo for the pull requests job to request python3 labels

PR #441, opened 2017-06-08
remove custom cluster names for deploying

alfredodeza:
It still uses 'ceph' everywhere but disables this customization 

http://tracker.ceph.com/issues/20228

### DIFF
--- a/ceph_deploy/cli.py
+++ b/ceph_deploy/cli.py
@@ -69,12 +69,6 @@ def get_parser():
         help='overwrite an existing conf file on remote host (if present)',
         )
     parser.add_argument(
-        '--cluster',
-        metavar='NAME',
-        help='name of the cluster',
-        type=validate.alphanumeric,
-        )
-    parser.add_argument(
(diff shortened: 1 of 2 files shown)


jammerhund:
To remove the --cluster option is not the best choise. In situations of multiple clusters in a single node you are right. But I try to setup a second cluster on new/other nodes. In this case the ceph-deploy tool is completely useless :-( Especially when you try to maintain multiple different clusters from one admin node. A better approach is to warn or ommit the usage of ceph-deploy in multiple clusters in a single node situations and to allow it for clusters on really different nodes.

alfredodeza:
@jammerhund please see ceph-devel discussion [0] - this is not only ceph-deploy. Ceph as a whole will no longer support custom cluster names.

If you need to deploy/manage multiple clusters you can still do that with ceph-deploy, you just need a different directory from where to operate from.

[0] http://lists.ceph.com/pipermail/ceph-users-ceph.com/2017-November/022244.html

PR #446, opened 2017-07-09
ceph-deploy to support on Debian 9 stretch

yaguangtang:
Adds Debian 9 stretch support


alfredodeza:
hi @yaguangtang this looks good to me. A couple of things:

* can you open a new issue in the tracker so that we can reference it in the commit (explained in the contributing guidelines https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#commits )

* Since Debian 10 has been defined as well, can you add that one too?

Thanks again!

PR #451, opened 2017-08-23
Specify partiton to ceph-disk for bluestore

zenglg:
usage: ceph-deploy osd prepare --bluestore <node>:<obj metadata>:<bluestore block>

Signed-off-by: Linggang Zeng <linggang.zeng@easystack.cn>


alfredodeza:
OK to test

alfredodeza:
@zenglg not sure how this works with ceph-disk, can you explain a bit more what ceph-disk expects here and how this wasn't working before but with this PR would?

Please review the contributing guidelines to open up an issue first https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#contributing-to-ceph-deploy

Maybe that is where a thorough explanation can go into what we are fixing here

zenglg:
@alfredodeza Sorry for the delay, I missed the confirmed E-mail from http://tracker.ceph.com until I found it in the trash today.
Add the RM-XXXX and repushed.
@yangdongsheng For the notification.

zenglg:
@alfredodeza  Ping

alfredodeza:
@zenglg I am still not sure about how this is supposed to work with ceph-disk, I am unclear what exactly are the options that ceph-disk wants/expects and how this wasn't working before

Could you please expand the ticket or this PR to include that information?

Thanks for amending the commit!

ceph_deploy/osd.py
```diff
@@ -227,6 +228,11 @@ def prepare_disk(
     if journal is not None:
         args.append(journal)
 
+    if block is not None:
+        # The null string is necessary for 'ceph-disk'.
```
alfredodeza: so this is a requirement from ceph-disk? See this is what my concern is, I had no idea that ceph-disk now requires things like this. Any background information explaining that would be great here

ceph_deploy/osd.py
```diff
@@ -322,8 +328,11 @@ def prepare(args, cfg, activate_prepared_disk):
                       hostname, disk, journal, activate_prepared_disk)
 
             storetype = None
+            block = None
             if args.bluestore:
                 storetype = 'bluestore'
+                block = journal
```
vasukulkarni: @zenglg I think you want to explain bluestore expects BLOCK (as compared to JOURNAL in filestore) and append that coment before this so as to avoid confusion.

ceph_deploy/osd.py
vasukulkarni: @zenglg If you can add the comments before and address the difference, otherwise LGTM

alfredodeza:
@zenglg since version 2.0.0 of ceph-deploy we are using ceph-volume on the remote nodes which takes care of this. There is no plant to support backported features like this for ceph-deploy. Are you OK with closing this or do you want to proceed in a different way?

PR #453, opened 2017-09-06
Use /etc/os-release if platform.linux_distribution() empty

tserong:
On systems with only /etc/os-release (rather than distro-specific
release files), platform.linux_distribution() returns ('', '', '').
In this case, fall back to a trivial parsing of /etc/os-release.

It would be cleaner to use the new distro.linux_distribution()
function, especially since platform.linux_distribution() is
deprecated, but I don't know that there's a way of ensuring that
module is present in advance on all the remote hosts.

Fixes: http://tracker.ceph.com/issues/18164
Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -8,12 +8,15 @@ import os
 import shutil
 import tempfile
 import platform
+import re
 
 
 def platform_information(_linux_distribution=None):
     """ detect platform information from remote host """
     linux_distribution = _linux_distribution or platform.linux_distribution
     distro, release, codename = linux_distribution()
+    if not distro:
+        distro, release, codename = parse_os_release()
     if not codename and 'debian' in distro.lower():  # this could be an empty string in Debian
         debian_codenames = {
             '10': 'buster',
@@ -44,6 +47,34 @@ def platform_information(_linux_distribution=None):
         str(codename).rstrip()
     )
 
+
+def parse_os_release(release_path='/etc/os-release'):
+    """ Extract (distro, release, codename) from /etc/os-release if present """
+    release_info = {}
+    if os.path.isfile(release_path):
+        for line in open(release_path, 'r').readlines():
+            line = line.strip()
+            if line.startswith('#'):
+                continue
+            parts = line.split('=')
+            if len(parts) != 2:
+                continue
+            release_info[parts[0].strip()] = parts[1].strip("\"'\n\t ")
+    # In theory, we want ID/NAME, VERSION_ID and VERSION_CODENAME (with a
+    # possible fallback to VERSION on the latter), based on information at:
+    # https://www.freedesktop.org/software/systemd/man/os-release.html
+    # However, after reviewing several distros /etc/os-release, getting
+    # the codename is a bit of a mess.  It's usually in parentheses in
+    # VERSION, with some exceptions.
+    distro = release_info.get('ID', '')
+    release = release_info.get('VERSION_ID', '')
+    codename = release_info.get('UBUNTU_CODENAME', release_info.get('VERSION', ''))
+    match = re.match(r'^[^(]+ \(([^)]+)\)', codename)
+    if match:
+        codename = match.group(1).lower()
+    if not codename and release_info.get('NAME', '') == 'openSUSE Tumbleweed':
+        codename = 'tumbleweed'
+    return (distro, release, codename)
 
 def machine_type():
     """ detect machine type """
--- a/ceph_deploy/hosts/__init__.py
+++ b/ceph_deploy/hosts/__init__.py
@@ -114,7 +114,7 @@ def _normalized_distro_name(distro):
         return 'scientific'
     elif distro.startswith('oracle'):
         return 'oracle'
-    elif distro.startswith(('suse', 'opensuse')):
+    elif distro.startswith(('suse', 'opensuse', 'sles')):
         return 'suse'
     elif distro.startswith('centos'):
         return 'centos'
--- a/ceph_deploy/tests/test_remotes.py
+++ b/ceph_deploy/tests/test_remotes.py
@@ -1,6 +1,6 @@
 from mock import patch
 from ceph_deploy.hosts import remotes
-from ceph_deploy.hosts.remotes import platform_information
+from ceph_deploy.hosts.remotes import platform_information, parse_os_release
 
 class FakeExists(object):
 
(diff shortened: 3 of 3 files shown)


tserong:
*ahem* wait, I've got a broken test

tserong:
...that's better.

ceph_deploy/hosts/remotes.py
```diff
@@ -14,6 +14,15 @@ def platform_information(_linux_distribution=None):
     """ detect platform information from remote host """
     linux_distribution = _linux_distribution or platform.linux_distribution
     distro, release, codename = linux_distribution()
+    if not distro and os.path.isfile('/etc/os-release'):
```
alfredodeza: I think this should be considered the most canonical way of doing this, and the `if not distro` should be the fallback

ceph_deploy/hosts/remotes.py
```diff
@@ -14,6 +14,15 @@ def platform_information(_linux_distribution=None):
...
+        for line in open('/etc/os-release', 'r').readlines():
+            parts = line.split('=')
```
alfredodeza: I would parse everything that exists and make it a dictionary, that way there is no need for all these `if` conditions, what was parsed is what you will get. It makes the parsing more robust as well, less prone to breakage if anything doesn't quite match.

There aren't any unit tests either for this, would love to see a few of them to back this new distro detection

ceph_deploy/hosts/remotes.py
tserong: Just to clarify, we should check /etc/os-release *first*, then fallback to platform.linux_distribution() if that doesn't exist?

ceph_deploy/hosts/remotes.py
tserong: OK, will do.

ceph_deploy/hosts/remotes.py
alfredodeza: yes, that sounds correct

tserong:
Looking at the existing tests in ceph_deploy/tests/test_remotes.py, ISTM that test_handles_no_codename(), test_hanles_centos_64() and test_handles_ubuntu_percise() are pretty much NOOPs. The tests themselves provide the expected results, then the platform_information() function populates (distro, release, codename) with what was passed in, and returns it. Or am I missing something?

alfredodeza:
@tserong those tests look valid to me. They are a bit deceiving because it looks like it just returns what gets passed in, but that isn't always the case, hence the tests. For example a `('debian', '8.4', '')` will correctly return 'jessie' for the codename.

The CentOS64 test is needed because we need to make sure those aren't changed, that is, those tests ensure that whatever happens to the code, the behavior for CentOS stays the same.

Since the code is opening `/etc/os-release` in side the function, it will get tricky to test it out easily. I would suggest extracting that functionality into a separate function (`def parse_os_release(path='/etc/os-release')`), that way the path can be changed to anything a test can write to (py.test can help here with `tmpdir` https://docs.pytest.org/en/latest/tmpdir.html )

So a test could look like:

    def test_os_release_parses_version(self, tmpdir):
        path = str(tmpdir)
        with open(path, 'w') as os_release:
            os_release.write("""
            VERSION="16.04 LTS (Xenial Xerus)"
            """)
        result = parse_os_relase(path)
        assert resul['VERSION'] == "16.04 LTS (Xenial Xerus)"

tserong:
a9b75d3 should be a bit better, although still uses os-release as the fallback, not the default. I was concerned about not having access to a sufficiently broad range of /etc/os-release files to use as a basis for writing tests :-/

ceph_deploy/hosts/remotes.py
```diff
@@ -45,6 +49,25 @@ def platform_information(_linux_distribution=None):
...
+                continue
+            parts = line.split('=')
+            if len(parts) != 2:
+                continue
+            release_info[parts[0].strip()] = parts[1].strip("\"'\n\t ")
+    # The below should be about right, per
+    # https://www.freedesktop.org/software/systemd/man/os-release.html,
+    distro = release_info.get('NAME', '')
+    release = release_info.get('VERSION_ID', '')
+    codename = release_info.get('VERSION_CODENAME', release_info.get('VERSION', ''))
```
alfredodeza: This would not work with Ubuntu (`codename` would return `''`):

    $ cat /etc/os-release
    NAME="Ubuntu"
    VERSION="16.04 LTS (Xenial Xerus)"
    ID=ubuntu
    ID_LIKE=debian
    PRETTY_NAME="Ubuntu 16.04 LTS"
    VERSION_ID="16.04"
    HOME_URL="http://www.ubuntu.com/"
    SUPPORT_URL="http://help.ubuntu.com/"
    BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
    UBUNTU_CODENAME=xenial

tserong: It tries VERSION_CODENAME first, then falls back to VERSION, so in this case it'll return `codename == '16.04 LTS (Xenial Xerus)'`. But yeah, we'd want it to return `'xenial'` in this case.

vasukulkarni:
It is surprising that no one has patched platform.linux_distribution() for suse yet, I believe you should also submit a patch there and probably remove this here once fixed.  Can you paste logs from your runs here for record?

vasukulkarni:
jenkins test this please

vasukulkarni:
LGTM

tserong:
Python's platform.linux_distribution() is deprecated and is being removed in (IIRC) python 3.7. Consequently nobody's ever going to update platform.linux_distribution() to support parsing /etc/os-release (see https://bugs.python.org/issue1322)...

tserong:
For the record, `ceph-deploy new` fails with:

````
[ceph_deploy][ERROR ] UnsupportedPlatform: Platform is not supported:   
````
i.e. it prints no platform information at all in the log, because it's `('', '', '')` on systems which have only /etc/os-release, but no other distro-specific release files or lsb stuff installed.

alfredodeza:
We really need to have Ubuntu support for this. Once that gets updated (and a test added) this is good to go

vasukulkarni:
@alfredodeza the platform.linux_distribution() works for ubuntu so this code path will not run and will be run only for suse.

wjwithagen:
On 12-9-2017 04:12, Tim Serong wrote:
> For the record, |ceph-deploy new| fails with:
> 
> |[ceph_deploy][ERROR ] UnsupportedPlatform: Platform is not supported: |
> 
> i.e. it prints no platform information at all in the log, because it's
> |('', '', '')| on systems which have only /etc/os-release, but no other
> distro-specific release files or lsb stuff installed.


FreeBSD would generate the same problem. It ain't no Linux.
But I haven't gotten around to ceph-deploy.

--WjW

alfredodeza:
@vasukulkarni the fix here is not for suse, this is a long standing issue because `linux_distribution()` is no longer a thing in Python 3 *and* `/etc/os-release` exists where the former fails.

vasukulkarni:
we have python 3 tests using ceph-deploy, I am not sure how its working there!! https://github.com/ceph/ceph/blob/master/qa/tasks/ceph_deploy.py#L56 

If its somehow broken for pure python 3, then the current /etc/ parsing for all is barely sufficient, an alternation to python_distribution is needed.

vasukulkarni:
I think this is going to be an issue from 3.7 onwards but for 3.5 it still works

`
Python 3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330] on linux
>>> from platform import dist, linux_distribution                                                                                                                            
>>> linux_distribution()                                                                                                                                                     
('Ubuntu', '16.04', 'xenial')
`

tserong:
> If its somehow broken for pure python 3, then the current /etc/ parsing for all is barely sufficient, an alternate to python_distribution is needed.

Best would be to use distro.linux_distribution() (from https://pypi.python.org/pypi/distro), but then we somehow have to make sure that's actually installed on all the remote hosts first, which is a bit of a chicken and egg problem.

tserong:
Now includes somewhat sane codename determination and tests for os-release on CentOS 7, Debian Stretch, Fedora 26, openSUSE Leap 42.2, openSUSE Tumbleweed and Ubuntu Xenial.

alfredodeza:
@tserong this is great, thank you so much for updating it to accommodate the other distros *and* all the tests!

tserong:
Cheers :-)

PR #454, opened 2017-09-12
remove python-distribute dependency

vasukulkarni:
This patch is straight from http://tracker.ceph.com/issues/16399, seems like its hit with openstack mitaka repo.

python-distribute was merged back into python-setuptools, so there should be no reason to require both.


alfredodeza:
@vasukulkarni can you reword the commit so that it has the convention as described in https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst#commits

This looks good to go once that gets done

vasukulkarni:
@alfredodeza fixed.

bryanapperson:
This still seems to be affecting the luminous RPMs on the ceph downloads portal.

alfredodeza:
@bryanapperson this is not yet released

PR #458, opened 2018-01-10
mgr recursively create folders

psukys:
If `/var/lib/ceph/mgr` directory is missing, `safe_mkdir` still does not fail, but subsequent code is unable to reach `/var/lib/ceph/mgr/cluster-hostname` directory, which will not be created (as `mgr` folder is missing)

### DIFF
--- a/ceph_deploy/mgr.py
+++ b/ceph_deploy/mgr.py
@@ -32,7 +32,7 @@ def create_mgr(distro, name, cluster, init):
         name=name
         )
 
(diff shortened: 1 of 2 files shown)


alfredodeza:
Could you open a ticket in the tracker and amend your commit? Basically what is described here: https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst

psukys:
@alfredodeza , will do a bit later and rebase the commit to comply

PR #459, opened 2018-01-16
[RM-22712] Use recursive dir creation

psukys:
If `/var/lib/ceph/mgr` directory is missing, `safe_mkdir` still does not fail, but subsequent code is unable to reach `/var/lib/ceph/mgr/cluster-hostname` directory, which will not be created (as `mgr` folder is missing)

This is formatted under contribution guidelines, as noted in #458.

Link to redmine issue: [22712](http://tracker.ceph.com/issues/22712)


alfredodeza:
ok to test

PR #461, opened 2018-01-26
RM-7505 Add support for Arch Linux

HacKanCuBa:
With the help of a friend, we added support for Arch Linux.
We tested it several times on different machines and VMs to make sure it works as expected.
I was specially careful on not changing current functionality.

### DIFF
--- a/ceph_deploy/hosts/arch/uninstall.py
+++ b/ceph_deploy/hosts/arch/uninstall.py
@@ -1,0 +1,50 @@
+import logging
+
+from ceph_deploy.util.system import disable_service, stop_service
+from ceph_deploy.lib import remoto
+
+
+SYSTEMD_UNITS = [
+    'ceph-mds.target',
+    'ceph-mon.target',
+    'ceph-osd.target',
+    'ceph-radosgw.target',
+    'ceph-fuse.target',
+    'ceph-mgr.target',
+    'ceph-rbd-mirror.target',
+    'ceph.target',
+]
+
+
+def uninstall(distro, purge=False):
+    packages = [
+        'ceph',
+    ]
+
+    hostname = distro.conn.hostname
+    LOG = logging.getLogger(hostname)
+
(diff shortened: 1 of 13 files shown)


ceph_deploy/hosts/arch/uninstall.py
```diff
@@ -0,0 +1,50 @@
...
+
+def uninstall(distro, purge=False):
+    packages = [
+        'ceph',
+    ]
+
+    hostname = distro.conn.hostname
+    LOG = logging.getLogger(hostname)
+
+    # I need to stop and disable services prior package removal
```
alfredodeza: this reads like a packaging problem on Arch's side. It is the package itself that should have rules that take care of this. I guess this doesn't happen here because you have encountered this?

HacKanCuBa: Exactly, Arch packages don't start nor stop services, neither it enables them.

HacKanCuBa:
BTW, I can help you with the testing and maintaining of this for Arch Linux, I have a productive Arch Linux server so I will use this for a long time.

ceph_deploy/hosts/arch/uninstall.py
HacKanCuBa: Would you prefer to notify the user, instead of starting and enabling services?

vasukulkarni:
No objections as well, but It will be nice to throw Warning message during install, something like "Arch is not a supported distro and is for experimental use" or something ( can be added in install.py).

We can merge this after that.

alfredodeza:
@vasukulkarni we don't do that for all the other distros that we don't support, so I am -1 on adding warnings just for Arch.

vasukulkarni:
@alfredodeza yep that is fair!, we have other distro's without warning message,  yet some point we should add the warning based on ceph tested distro's but is a different issue.

PR #462, opened 2018-01-26
[RM-22816] Fix osd.py:disk_list

HacKanCuBa:


alfredodeza:
OK to test

westsouthnight:
Anybody can fix that?

ceph_deploy/osd.py
```diff
@@ -372,8 +372,8 @@ def disk_list(args, cfg):
             command,
         )
         for line in out:
-            if line.startswith('Disk /'):
-                distro.conn.logger(line)
+            if line.startswith(b'Disk /'):
```
alfredodeza: this is somewhat correct, I think the better approach would be to check if the incoming string is a bytearray or a plain string. A utility that can be re-used would be better, as this is not the only place in ceph-deploy where this check would be needed. Something like:

    if isinstance(string, bytes):
        # we really ignore here if we can't properly decode with utf-8
        return string.decode('utf-8', 'ignore')
    return string

alfredodeza: *maybe* even better would be to make remoto do this automatically for us, optionally turning it off (but decoding to string by default)

HacKanCuBa: That would be great, and the actual right way of doing it. I didn't want to send a whole refactor patch :P
Given that `remoto` is your lib, you should decide whether it's gonna be done by it or by this in an aux function. How would you like to proceed?

alfredodeza: Lets get this done on the remoto side, I'll take care of the packaging. Would you like to try a PR against remoto?

HacKanCuBa: I had a heavy week, and will have a heavy one, sorry for the delay.
But yeah, lets get on it. I need to catch up on the code, so probably it's gonna be for the end of the week or so.

HacKanCuBa:
I'm clossing this since the solution will be applied on the `remoto` lib.

LinBoLen:
ceph-deploy 2.0.1 still complain error 
```
[INFO  ] Running command: sudo fdisk -l
[ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/util/decorators.py", line 69, in newfunc
[ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/cli.py", line 166, in _main
[ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/osd.py", line 434, in disk
[ceph_deploy][ERROR ]     disk_list(args, cfg)
[ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/osd.py", line 375, in disk_list
[ceph_deploy][ERROR ]     if line.startswith('Disk /'):
[ceph_deploy][ERROR ] TypeError: startswith first arg must be bytes or a tuple of bytes, not str
[ceph_deploy][ERROR ]
```

HacKanCuBa:
> ceph-deploy 2.0.1 still complain error
> 
> ```
> [INFO  ] Running command: sudo fdisk -l
> [ceph_deploy][ERROR ] Traceback (most recent call last):
> [ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/util/decorators.py", line 69, in newfunc
> [ceph_deploy][ERROR ]     return f(*a, **kw)
> [ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/cli.py", line 166, in _main
> [ceph_deploy][ERROR ]     return args.func(args)
> [ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/osd.py", line 434, in disk
> [ceph_deploy][ERROR ]     disk_list(args, cfg)
> [ceph_deploy][ERROR ]   File "/usr/local/lib/python3.7/site-packages/ceph_deploy/osd.py", line 375, in disk_list
> [ceph_deploy][ERROR ]     if line.startswith('Disk /'):
> [ceph_deploy][ERROR ] TypeError: startswith first arg must be bytes or a tuple of bytes, not str
> [ceph_deploy][ERROR ]
> ```

hmmm weird, can you give us info about OS and stuff?

HacKanCuBa:
Wait, right, my bad. There's no new release of ceph-deploy! Run it by pulling from master, manually, instead of using your package manager. Because this is already fixed except there has been no new release, I don't know why.

PR #463, opened 2018-01-29
net: Return both IPv4 and IPv6 addresses

wido:
When using ceph-deploy it will now also show the IPv6 addresses
it found for a node:

<pre>IP addresses found: [u'192.168.153.10', u'2001:db8::100', u'10.0.2.15']</pre>

The resulting ceph.conf will use IPv6 addresses for the Monitors:

<pre>
ms_bind_ipv6 = true
mon_host = [2001:db8::100],[2001:db8::101],[2001:db8::102]
</pre>

### DIFF
--- a/ceph_deploy/util/net.py
+++ b/ceph_deploy/util/net.py
@@ -70,14 +70,14 @@ def in_subnet(cidr, addrs=None):
 
 def ip_addresses(conn, interface=None, include_loopback=False):
     """
-    Returns a list of IPv4 addresses assigned to the host. 127.0.0.1 is
+    Returns a list of IPv4/IPv6 addresses assigned to the host. 127.0.0.1/::1 is
     ignored, unless 'include_loopback=True' is indicated. If 'interface' is
     provided, then only IP addresses from that interface will be returned.
 
     Example output looks like::
 
         >>> ip_addresses(conn)
-        >>> ['192.168.1.111', '10.0.1.12']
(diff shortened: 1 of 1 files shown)


alfredodeza:
Looks OK, but needs some tests added

wido:
Thanks @alfredodeza 

Which file do you suggest? test_cli_new.py?

alfredodeza:
I just realized this particular function doesn't have any tests at all :( If I were to add anything it would be to `tests/unit/util/test_net.py`

Might be a bit tricky to get it right, hopefully you can add something to it. Thank you!

vasukulkarni:
I think this should also fix http://tracker.ceph.com/issues/21427 , OVH run would ensure that it works with IPv6, I will schedule some run today, Nice to see this PR!

@yuriw  FYI

wido:
@alfredodeza I'll look into a test for this function.

And thanks @vasukulkarni !

vasukulkarni:
this run failed for a different reason, but atleast the monitors came up with IPv4, this atleast tests for regression, so I am approving it.

http://qa-proxy.ceph.com/teuthology/vasu-2018-02-01_20:48:20-ceph-deploy-master-distro-basic-ovh/2138594/teuthology.log

wido:
Thanks! I can confirm that my MONs came up with IPv6 as they were supposed to :)

vasukulkarni:
@alfredodeza ping? anything pending?

PR #465, opened 2018-02-16
do not consider directories as executable paths

alfredodeza:
Fixes: http://tracker.ceph.com/issues/21677


alfredodeza:
test this please

PR #469, opened 2018-03-03
Fix UnboundLocalError when fail to get hostname

int32bit:
If we fail to get hostname, the `hosts.get` will raise an exception and
don't return any value, so the variable `distro` in [1] and [2] may be
undefined. What's worse, we may get previous value if we fail in second
time but succeed in first time.

In order to avoid the issues above, we should initialize the `distro`
value in each loop iteration.

[1] https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mgr.py#L173
[2] https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mds.py#L173

### DIFF
--- a/ceph_deploy/mds.py
+++ b/ceph_deploy/mds.py
@@ -139,6 +139,7 @@ def mds_create(args):
 
     for hostname, name in args.mds:
         try:
+            distro = None
             distro = hosts.get(hostname, username=args.username)
             rlogger = distro.conn.logger
             LOG.info(
@@ -170,7 +171,7 @@ def mds_create(args):
             create_mds(distro, name, args.cluster, distro.init)
             distro.conn.exit()
         except RuntimeError as e:
-            if distro.normalized_name == 'redhat':
+            if distro and distro.normalized_name == 'redhat':
                 LOG.error('this feature may not yet available for %s %s' % (distro.name, distro.release))
                 failed_on_rhel = True
             LOG.error(e)
--- a/ceph_deploy/mgr.py
+++ b/ceph_deploy/mgr.py
@@ -139,6 +139,7 @@ def mgr_create(args):
 
     for hostname, name in args.mgr:
         try:
+            distro = None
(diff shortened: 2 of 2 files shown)


int32bit:
The following is the error info that we fix before:

```
[root@int32bit-server-1 ceph-deploy(keystone_admin)]# ceph-deploy mds create test-1
[ceph_deploy.mds][DEBUG ] Deploying mds, cluster ceph hosts test-1:test-1
ssh: Could not resolve hostname test-1: Name or service not known
[ceph_deploy][ERROR ]   File "/usr/lib/python2.7/site-packages/ceph_deploy/util/decorators.py", line 69, in newfunc
[ceph_deploy][ERROR ]   File "/usr/lib/python2.7/site-packages/ceph_deploy/cli.py", line 164, in _main
[ceph_deploy][ERROR ]   File "/usr/lib/python2.7/site-packages/ceph_deploy/mds.py", line 191, in mds
[ceph_deploy][ERROR ]     mds_create(args)
[ceph_deploy][ERROR ]   File "/usr/lib/python2.7/site-packages/ceph_deploy/mds.py", line 173, in mds_create
[ceph_deploy][ERROR ]     if distro.normalized_name == 'redhat':
[ceph_deploy][ERROR ] UnboundLocalError: local variable 'distro' referenced before assignment
[ceph_deploy][ERROR ]
```

We should raise clearer error info to user rather than syntax error.

alfredodeza:
On what version of ceph-deploy did this fail?

Could you open an issue first so that we can follow up on your PR (see https://github.com/ceph/ceph-deploy/blob/master/CONTRIBUTING.rst )

int32bit:
Thanks @alfredodeza, I will create the issue first and recommit this in other thread.

alfredodeza:
You can ammend your commits and repush. No need to open another PR. But whatever you feel more comfortable with is fine with me

PR #470, opened 2018-03-05
[RM-23218] fix UnboundLocalError when create mds/mgr with bad hosts

int32bit:
If we fail to get hostname, the hosts.get will raise an exception and
don't return any value, so the variable distro in [1] and [2] may be
undefined. What's worse, we may get previous value if we fail in second
time but succeed in first time.

In order to avoid the issues above, we should initialize the distro
value in each loop iteration.

[1] https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mgr.py#L173
[2] https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mds.py#L173

### DIFF
--- a/ceph_deploy/mds.py
+++ b/ceph_deploy/mds.py
@@ -139,6 +139,7 @@ def mds_create(args):
 
     for hostname, name in args.mds:
         try:
+            distro = None
             distro = hosts.get(hostname, username=args.username)
             rlogger = distro.conn.logger
             LOG.info(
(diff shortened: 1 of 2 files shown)


int32bit:
Fix http://tracker.ceph.com/issues/23218

alfredodeza:
@int32bit can you tell me what version of ceph-deploy are you using when this happens?

int32bit:
@alfredodeza I ran the code in master branch, and I'm sure it still exists in current code [1].

[1] https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/mgr.py#L142

alfredodeza:
@int32bit this looks good. Would you mind one more change? Can you split your 1 commit into one for each file? That way it makes it easier if we ever need to bisect/cherry-pick this

alfredodeza:
OK to test

int32bit:
Hi, @alfredodeza, the PR has been split into two commits, please review it again, thanks.

alfredodeza:
Thank you @int32bit !

PR #478, opened 2018-10-24
[RM-36583] align with create by using short hostname

iotcg:
If using fully qualified domain name to check monitor
status after "mon add", it will hit the error that can't
find the node file under /var/run/ceph/ which result
in aboring add monitor

Signed-off-by: Changcheng Liu <changcheng.liu@intel.com>

### DIFF
--- a/ceph_deploy/mon.py
+++ b/ceph_deploy/mon.py
@@ -193,6 +193,7 @@ def mon_add(args):
     )
 
     mon_section = 'mon.%s' % mon_host
+    (name, host) = next(mon_hosts(args.mon))
     cfg_mon_addr = cfg.safe_get(mon_section, 'mon addr')
 
     if args.address:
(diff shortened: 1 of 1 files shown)


iotcg:
When using "mon add" command, such as "ceph-deploy mon add nstcloudcc1.sh.intel.com",
It will check the monitor status with command " sudo ceph --cluster=ceph --admin-daemon /var/run/ceph/ceph-mon.nstcloudcc1.sh.intel.com.asok mon_status". However, the actual file name is "/var/run/ceph/ceph-mon.nstcloudcc1.asok". The reason is it use "fully qualified domain name" as part of the file name. Actually, it should use short hostname.

I've checked the "mon create" command, such as "ceph-deploy mon create nstcloudcc1.sh.intel.com", it use the short hostname to check the monitor status "sudo ceph --cluster=ceph --admin-daemon /var/run/ceph/ceph-mon.nstcloudcc1.asok mon_status" and succeed.

tchaikov (commented):
@changchengx could you split this commit into two? one for using `name` returned by `mon_hosts()`, another for transposing `mon_status()` nad `catch_mon_errors()`?

changchengx:
@tchaikov 
I'll check this PR within this week. It was filed about 2 years ago.

tchaikov:
> @tchaikov
> I'll check this PR within this week. It was filed about 2 years ago.

the original maintainer does not work on this project anymore. guess that's why your PR was unintended.

PR #479, opened 2018-10-25
Add ? after question

deshpandeshivam23:


HacKanCuBa:
It is not a question, but a statement. And if it were, it shouldn't use a space before the question mark.

tchaikov:
i concur with @HacKanCuBa.

PR #480, opened 2019-01-09
Fix remoto vendoring, bump remoto version to 0.0.35

alfredodeza:
A change in the remoto version caused vendoring not to work anymore. We now declare our dependencies directly in Python, spec file, and control file.

Bumps remoto to use the latest version available (0.0.35 at this time) which fixes the following error:

    NameError: global name 'extend_path' is not defined

Fixes: http://tracker.ceph.com/issues/37848

### DIFF
--- a/ceph-deploy.spec
+++ b/ceph-deploy.spec
@@ -31,8 +31,7 @@ BuildRequires:  pytest
 %endif
 BuildRequires:  git
 Requires:       python-argparse
-#Requires:      lsb-release
-#Requires:      ceph
+Requires:       python-remoto
 %if 0%{?suse_version} && 0%{?suse_version} <= 1110
 %{!?python_sitelib: %global python_sitelib %(python -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
 %else
--- a/debian/control
+++ b/debian/control
@@ -15,6 +15,7 @@ Architecture: all
 Depends: python,
          python-argparse,
          python-setuptools,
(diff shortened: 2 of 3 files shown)


alfredodeza:
This is going to require a new version of `remoto` (0.0.36). That version brings vendoring back. Current ceph-deploy will continue to be broken until this gets merged. See https://github.com/alfredodeza/remoto/pull/43

alfredodeza:
@vasukulkarni these changes should (for real!) fix your test runs now. Could you double check all is good?

@ktdreyer seems like the python-remoto version in EPEL is super old, is there anything I can help with to get `0.0.35` (https://github.com/alfredodeza/remoto/releases/tag/0.0.35) packaged+released?

vasukulkarni:
LGTM, going to run a simple test with teuthology using this branch.

vasukulkarni:
I can confirm that original bootstrap issue is fixed and this is ready for merge, I did quick test on my laptop. Also scheduled some tests in ovh but its going to take time, we dont have to wait for that http://pulpito.ceph.com/vasu-2019-01-15_16:41:00-ceph-deploy-master-distro-basic-ovh/

alfredodeza:
I'll go ahead and merge. Still worried about the old (version 0.0.25) python-remoto in EPEL, but that would affect an upcoming release of ceph-deploy. Will have to follow up with test fixes since pytest has dropped support for some helpers being used.

PR #490, opened 2019-09-23
Add support for EuroLinux

AlexBaranowski:
Signed-off-by: Alex Baranowski <alex@euro-linux.com>

This PR contains the patch that enables ceph-deploy on EuroLinux. I tested on both CentOS (in case of Ceph deploy EuroLinux uses CentOS like installation etc.) and EuroLinux.

Unfortunately on the **current master** some tests fail.

For Python 2.7:

- test_section_repeat()

Final result 
```
1 failed, 2027 passed, 16 skipped, 39 warnings in 17.20 seconds
```

For Python 3.6:

- test_executable_is_used(self)
- test_host_is_used(self)

Final result:
```
2 failed, 2026 passed, 16 skipped, 46 warnings in 15.82s
```

#### After patch

After the patch **the previous tests**, that are not connected with
the patch also fail, but the new tests work seamlessly, that means
that there is no regression.

I also change .gitignore, so PyCharm and .eggs files are ignored. 

I also noticed that there is half-finished CI on jenkins.ceph.com, I'm currently working on our in-house QA process, so if there is public repo, I could help with it.

### DIFF
--- a/ceph_deploy/hosts/centos/install.py
+++ b/ceph_deploy/hosts/centos/install.py
@@ -5,7 +5,6 @@ from ceph_deploy.hosts.common import map_components
 from ceph_deploy.util.paths import gpg
 from ceph_deploy.util import net
 
-
 LOG = logging.getLogger(__name__)
 NON_SPLIT_PACKAGES = ['ceph-osd', 'ceph-mon', 'ceph-mds']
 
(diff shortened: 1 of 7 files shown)


ceph_deploy/hosts/centos/install.py
```diff
@@ -5,13 +5,13 @@
...
 from ceph_deploy.util import net
 
-
 LOG = logging.getLogger(__name__)
 NON_SPLIT_PACKAGES = ['ceph-osd', 'ceph-mon', 'ceph-mds']
 
 
 def rpm_dist(distro):
-    if distro.normalized_name in ['redhat', 'centos', 'scientific', 'oracle', 'virtuozzo'] and distro.normalized_release.int_major >= 6:
+    if distro.normalized_name in ['redhat', 'centos', 'scientific', 'oracle', 'virtuozzo', 'eurolinux'] \
```
HacKanCuBa: Maybe we can take the opportunity to do something like this instead:

```python
names = {'redhat', 'centos', 'scientific', 'oracle', 'virtuozzo', 'eurolinux'}
if distro.normalized_name in names and distro.normalized_release.int_major >= 6:
```

Note: it's just a suggestion

ceph_deploy/hosts/centos/install.py
```diff
@@ -68,35 +68,18 @@ def install(distro, version_kind, version, adjust_repos, **kw):
...
+                'name': 'ceph-{suffix}'.format(suffix=repo_suffix),
+                'baseurl': 'https://repo.eurolinux.local/ceph/14_{suffix}/prod/'.format(suffix=repo_suffix),
+                'gpgcheck': '1',
+                # actually it's prefix but nvm
+                'gpgkey_url': 'https://repo.eurolinux.local/ceph/{suffix}_gpg.key'.format(suffix=repo_suffix),
+                'ssl_verify': '0'
+            }
+            yum_repo_text = '[{name}]\nname=Ceph {name}\nbaseurl={baseurl}' + \
+                            '\ngpgcheck={gpgcheck}\ngpgkey={gpgkey_url}\nsslverify={ssl_verify}'
+            distro.conn.remote_module.write_yum_repo(yum_repo_text)
```
HacKanCuBa: Wow, this section changed radically.
I'm not sure I understand correctly what's going on, but the lines removed added different repos for stable and testing, both hosted by *ceph*, then removed any preexisting version prior installing the new one.
The lines added doesn't do anything like this, but instead it seems that a new single repo is added (whether stable or testing) hosted by *eurolinux*, and there's no removal/installation of any package.

I'm not saying it's right or wrong, just that the functionality changed greatly.

AlexBaranowski: Sorry, but this one is wrong :). I forgot about this PR (since it was dead for like half year), and made some changes for our internal ceph-deploy. I will remove this PR.

AlexBaranowski:
Sorry for burning PR, but my current master don't match original one.

PR #496, opened 2020-05-22
Fix compatibility with Python 3.8

javacruft:
The deprecated platform.linux_distribution function was removed
in Python 3.8 - fallback to using parse_os_release if the function
is not found.

Signed-off-by: James Page <james.page@ubuntu.com>

### DIFF
--- a/ceph_deploy/hosts/remotes.py
+++ b/ceph_deploy/hosts/remotes.py
@@ -13,8 +13,13 @@ import re
 
 def platform_information(_linux_distribution=None):
     """ detect platform information from remote host """
-    linux_distribution = _linux_distribution or platform.linux_distribution
-    distro, release, codename = linux_distribution()
+    distro = release = codename = None
+    try:
+        linux_distribution = _linux_distribution or platform.linux_distribution
+        distro, release, codename = linux_distribution()
+    except AttributeError:
+        # NOTE: py38 does not have platform.linux_distribution
+        pass
     if not distro:
         distro, release, codename = parse_os_release()
     if not codename and 'debian' in distro.lower():  # this could be an empty string in Debian


ceph_deploy/hosts/remotes.py
```diff
@@ -13,8 +13,13 @@
 
 def platform_information(_linux_distribution=None):
     """ detect platform information from remote host """
-    linux_distribution = _linux_distribution or platform.linux_distribution
-    distro, release, codename = linux_distribution()
+    distro = release = codename = None
```
HacKanCuBa, line 16: maybe we can take this opportunity to simplify code a bit:

```python
    if _linux_distribution is not None:
        linux_distribution = _linux_distribution
    elif hasattr(platform, 'linux_distribution'):
        linux_distribution = platform.linux_distribution
    else:
        linux_distribution = parse_os_release
    distro, release, codename = linux_distribution()
```

changchengx:
+1

changchengx:
https://pypi.org/project/distro/ may help solve the problem.
I'm not sure whether it's also supported in python2.

tchaikov:
jenkins, retest this please

tchaikov:
jenkins, retest this please

tchaikov:
jenkins, retest this please

tchaikov:
the test failure should be addressed by #498.

HacKanCuBa:
> https://pypi.org/project/distro/ may help solve the problem.
> I'm not sure whether it's also supported in python2.

I looked at it before and seems OK, but I wouldn't add a whole dependency to replace just 4 LoC. I don't know if that package could be useful for other parts of the app that would make it valuable to require it.

tchaikov:
jenkins, retest this please

Kaizhe:
Looks like the release 2.0.1 doesn't include the fix? When do we expect have it in the new release ?

tchaikov:
> Looks like the release 2.0.1 doesn't include the fix? When do we expect have it in the new release ? As I still hit the issue:
> 

@Kaizhe v2.0.1 was cut 2 years ago, and it didn't include this change. see https://github.com/ceph/ceph-deploy/releases/tag/v2.0.1

Kaizhe:
Is there a plan to cut a new release?

tchaikov:
> Is there a plan to cut a new release?

i sent a mail to ceph-maintainers two months ago for inputs regarding to the interests in using ceph-deploy. no replies. so my understanding is that the community has no more interests in using it.

HacKanCuBa:
> > Is there a plan to cut a new release?
> 
> i sent a mail to ceph-maintainers two months ago for inputs regarding to the interests in using ceph-deploy. no replies. so my understanding is that the community has no more interests in using it.

Holly cr*p, this can't happen! I mean, besides using k8s, ceph-deploy is the easiest way to deploy ceph which otherwise is very complicated.
Which maintainers are you reaching? Debian, Fedora?

tchaikov:
it's a mailing list. please search for "ceph-maintainers" https://ceph.io/irc/. you might want to subscribe it for reading its archive.

zdover23:
I'm commenting here as a note to myself, because the potential removal of
ceph-deploy has significant documentation impact. (I'm the upstream Ceph
docs guy.) You don't need to read this email unless you have strong
opinions about the nature of the upstream Ceph documentation.

Currently, as of August 2020, the upstream ceph documentation contains
numerous references to ceph-deploy, even though as of the release of
Octopus (March-ish 2020) ceph-deploy is not the preferred method of
installing a Ceph cluster. The preferred method of installing Ceph is now
by using the cephadm tool.

I have begun to think that the best way to document the Ceph project is to
1) do our best to organize the grandma's attic of docs.ceph.com so that the
tree pane is more meaningful, and 2) create a new version of the docs each
time there is a release, and incrementally update that version only to
correct errors that were made when the document was originally composed.
Doing otherwise now would require the documentation to be changed
everywhere that ceph-deploy is mentioned, and those mentions would have to
be replaced with mentions of cephadm. If HacKan is correct and many people
still use ceph-deploy, then clearly it would be irresponsible and
destructive to remove the documentation that explains how to use
ceph-deploy.

Again, this has been mostly a note to self, but it's a topic that I have
been much aware of over the past month, and it's a thorny problem if we
don't make sure to archive old docs and create new docs for each release.

abalib:
> > Is there a plan to cut a new release?
> 
> i sent a mail to ceph-maintainers two months ago for inputs regarding to the interests in using ceph-deploy. no replies. so my understanding is that the community has no more interests in using it.

Ubuntu 20.04.3 LTS ships ceph-deploy and includes this bug.  It's quite a surprise ceph-deploy arrives broken out of the box.

mysiki:
Same, from `pip index versions ceph-deploy` I show than version 2.1.0 is not present in pip repo ... Can you push it ?

comay:
Just wanted to reiterate that it would be immensely helpful if (the final?) release of `2.1.0` was pushed to PyPi.

PR #500, opened 2020-08-06
setup: Install configparser when running on py2

badone:
Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,9 @@ install_requires = ['remoto>=1.1.4']
 pyversion = sys.version_info[:2]
 if pyversion < (2, 7) or (3, 0) <= pyversion <= (3, 1):
     install_requires.append('argparse')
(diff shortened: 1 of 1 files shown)


badone:
Nautilus ceph-deploy tests use the './bootstrap 2' call which a7820a272b19b79e7317ed8564a25019f2f7ce0d appears to have broken causing the following error.

```
stderr:[ceph_deploy][ERROR ] ImportError: No module named configparser
```

badone:
I think this was potentially missed because when you install the requirements for testing under py2 as described in requirements-dev.txt 'configparser' gets pulled in as a dependency and the tests pass. I notices when this is the case it installs configparser 4.0.2 whereas when installed using my patch it installs 5.0.0. I'm happy to change my patch to pin the version to 4.0.2 if we think that is the best strategy?

PR #507, opened 2020-10-01
bump to 2.1.0

tchaikov:
Signed-off-by: Kefu Chai <kchai@redhat.com>


comay:
Would it be possible to have this version pushed up to PyPi? @tchaikov @alfredodeza

